PR #29, opened 2026-05-25
fix: release STATE_LOCK during init_state OpenRouter ground-truth HTTP

fxdv:
## Summary
`init_state()` in `supervisor/state.py` calls `check_openrouter_ground_truth()` (OpenRouter HTTP) while holding `STATE_LOCK`. This blocks all state I/O for up to ~10s at every Colab boot and is the same class of regression that v6.0.0 fixed for `update_budget_from_usage`. This PR mirrors that fix in `init_state`.

## Change
- Release `STATE_LOCK` around the OpenRouter HTTP request inside `init_state` (`supervisor/state.py`).
- Re-acquire the lock to apply the result, re-reading any concurrently-mutable state.
- Regression test asserts the lock is not held during the HTTP call.

## BIBLE alignment
- **P1 Continuity** — eliminates a ~10s startup stall that delays state I/O for every other consumer.
- **P5 Minimalism** — pattern already exists in the same file (`update_budget_from_usage`); diff is small.
- **P8 Iteration** — one coherent change.

## Test plan
- [x] New regression test fails against `main`, passes against this branch.
- [x] `make test` passes locally (132 passed: 129 baseline + 3 new).
- [x] `make lint` clean for changed files (3 pre-existing ruff errors in `supervisor/state.py` are unrelated and untouched, per P5).

## Notes
- VERSION / `pyproject.toml` / README version intentionally **not** bumped — per BIBLE P7 the triple sync is the maintainer's prerogative on release.
- No changes to public API, no new dependencies, no touched abstractions.

### DIFF
--- a/supervisor/state.py
+++ b/supervisor/state.py
@@ -211,25 +211,27 @@ def init_state() -> Dict[str, Any]:
     Fetches OpenRouter ground truth and stores session_daily_snapshot and
     session_spent_snapshot for drift calculation.
     """
+    # Step 1: HTTP to OpenRouter OUTSIDE the lock (can take up to 10s).
+    # Mirrors v6.0.0 fix in update_budget_from_usage — never block STATE_LOCK on HTTP.
+    ground_truth = check_openrouter_ground_truth()
+
+    # Step 2: Apply snapshots and ground truth under the lock.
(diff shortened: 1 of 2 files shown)


fxdv:
closed due to deprecation of target

PR #42, opened 2026-06-07
fix(chat-ui): chronological card insertion + scope composer hover to pointer devices

ndrew1337:
Two UI fixes from a mobile bug handoff (BUG-2 is subsumed by BUG-3).

## BUG-1 — Consilium button looks "stuck red" on iOS Safari
The second tap **does** disarm it (`data-armed="false"`); the leftover red is a sticky `:hover`. iOS Safari keeps `:hover` after a tap, and the composer hover rules weren't gated.
**Fix (`web/style.css`):** wrap `.chat-consilium:hover`, `.chat-context-mode .chat-seg:hover`, and `.chat-send-inline:hover` in `@media (hover: hover)` so hover styling applies only to pointer devices. The armed state (fill + glow ring) stays visually distinct from hover.

## BUG-3 — chat cards/messages not inserted chronologically (fixes BUG-2)
On load/reconnect, task/progress/skill-lifecycle cards and the finished **background-consciousness** card appear at the **bottom** instead of their historical position; "yesterday's" finished cards re-materialise at the bottom on every reload.

Root cause is frontend-only: `/api/chat/history` already returns history sorted by `ts`, but `insertMessageNode` never ordered by time, and several paths (terminal-status loop, disconnected-card fallback, `task_summary`-only cards) bypass the ordered walk and append at the bottom.

**Fix (`web/modules/chat.js`):** stamp every bubble and live-card root with a sortable epoch-ms `data-ts` and make `insertMessageNode` insert chronologically.
- `data-ts` is derived from the **raw** timestamp (`Date.parse`), **never** from `normalizeLogTs()` — verified to be a lossy, time-only display string that can't order across days.
- Card roots are stamped in `applyLiveCardState` (the convergence funnel) with the raw ts threaded through `queueTaskLiveUpdate`/`bufferLiveUpdate`; **set-once/`Math.min`** anchors a card at its start so updates don't re-sort it to the bottom.
- `insertMessageNode` scans only direct children, **skips** children without a parseable `data-ts` (welcome/ephemeral/typing) rather than coercing to `0`/`NaN`, keeps the typing-indicator special-case and sticky-scroll, inserts before the first greater `data-ts`, else appends (nodes lacking `data-ts` still append — live streaming not regressed).
- Reused (`bg-consciousness`) cards drop their anchor on recycle so a new cycle re-stamps; subagent child cards live in their own container and are intentionally not ts-ordered.

**BUG-2 for free:** the finished bg-consciousness card carries the bg entry's `ts` and now settles into its historical position on reload — no replay stripping in `history.py`/`log_events.js`.

## Tests / verification
Adds `tests/test_chat_chronological_order.py` (source contracts for the load-bearing pieces). The behavioural ordered-insertion assertion belongs in the Playwright `ui_browser` lane — this is a host-renderer change and should be verified visually (desktop hover still highlights; iOS tap toggles cleanly; reload places cards chronologically) before release. `GATEWAY_CONTRACT_VERSION` == `VERSION` parity preserved.

### DIFF
--- a/tests/test_chat_chronological_order.py
+++ b/tests/test_chat_chronological_order.py
@@ -1,0 +1,67 @@
+"""Static UI contracts for the mobile/chat fixes (see MOBILE_BUGS.md).
+
+BUG-1: composer :hover styling is scoped to pointer devices so iOS Safari's
+sticky :hover does not leave the Consilium button looking "still on" after a tap.
+
+BUG-3 (subsumes BUG-2): chat bubbles and live-card roots are stamped with a
+sortable epoch-ms data-ts derived from the RAW timestamp (not the lossy
+normalizeLogTs display string), and insertMessageNode inserts in chronological
+order — so replayed/finished cards (including the background-consciousness card)
+settle into their historical position instead of being pinned to the bottom.
+
+The behavioural assertion lives in the Playwright ui_browser lane; these are
+source-contract pins for the load-bearing pieces.
+"""
+
+from __future__ import annotations
+
+import pathlib
+
+REPO = pathlib.Path(__file__).resolve().parents[1]
+
+
+def _read(rel: str) -> str:
+    return (REPO / rel).read_text(encoding="utf-8")
+
+
+# ───────────────────────── BUG-1: sticky hover on touch ─────────────────────
+
+def test_composer_hover_scoped_to_pointer_devices():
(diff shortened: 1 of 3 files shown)


ndrew1337:
Closing for now per author — will revisit. Branch kept so it can be reopened.

PR #49, opened 2026-06-13
fix(chat): owner "stop"/messages reach the running task + survive restart

ndrew1337:
## Problem

While a task is **busy**, owner messages typed into web chat (e.g. `стоп` / `остановись`, or any
mid-task instruction) did **not** reach the running task. The agent kept working and never reacted
to "stop". Pending owner messages were also **lost across a restart**.

## Root cause

1. **`server.py` busy-branch** routed any web-chat envelope as a *re-plan / new-task constraint*:
   `if task_constraint or task_metadata:`. Web chat **always** attaches a `task_metadata` envelope,
   so every busy-time owner message was misclassified as a constraint instead of being injected as
   plain owner text into the live task's mailbox.
2. **`owner_mailbox.write_owner_message`** swallowed I/O errors and returned `None`, so callers
   could believe a message was persisted when it wasn't.
3. **`supervisor/workers.py` `auto_resume_after_restart`** didn't drain orphaned owner messages, so
   anything queued while busy was dropped on the next restart.

## Fix

- **`server.py`** — busy-branch guard only treats the message as a constraint when
  `task_metadata.get("force_plan")` is set; otherwise it persists the plain owner text to the live
  task's mailbox via `write_owner_message(..., task_id=busy_task_id)`, falling back to
  `inject_message` when persistence returns `False` (or for images).
- **`ouroboros/gateway/ws.py`** — chat frame only attaches `task_metadata={"force_plan": True, ...}`
  when `force_plan` is truthy; plain messages carry no envelope, so they route as owner text.
- **`ouroboros/owner_mailbox.py`** — `write_owner_message` now returns `bool` (True iff appended;
  False on I/O error) so the caller can branch on real persistence.
- **`supervisor/workers.py`** — `auto_resume_after_restart` drains orphaned owner messages into the
  resume prompt (cap 20), resumes on an empty scratchpad when messages are pending, re-checks panic
  after the init sleep, and purges mailboxes on the panic / owner-restart branches so suppressed
  messages can't resurrect on a later restart.

## Tests

- `tests/test_busy_message_persistence.py` (new) — busy-time owner text persists to the live task's
  mailbox; `force_plan` still routes as a constraint; I/O-failure falls back to `inject_message`.
- `tests/test_loop_misc.py` — auto-resume drains orphaned owner messages; panic/owner-restart purge.

### DIFF
--- a/ouroboros/gateway/ws.py
+++ b/ouroboros/gateway/ws.py
@@ -264,18 +264,25 @@ async def ws_endpoint(websocket: WebSocket) -> None:
                         image_b64, image_mime, image_caption = _first_image_attachment(
                             msg.get("attachments")
                         )
-                        bridge.ui_send(
-                            payload,
+                        send_kwargs = dict(
                             sender_session_id=str(msg.get("sender_session_id", "") or ""),
                             client_message_id=str(msg.get("client_message_id", "") or ""),
                             image_base64=image_b64,
                             image_mime=image_mime,
                             image_caption=image_caption,
-                            task_metadata={
-                                "force_plan": force_plan,
-                                "force_plan_source": "consilium" if force_plan else "",
-                            },
                         )
+                        # Only attach planning metadata when a plan is actually
+                        # forced. A plain message must carry NO task_metadata so
+                        # that, if it arrives while a task is running, it is
+                        # delivered INTO that task (owner_mailbox) rather than
+                        # spawned as a separate queued task. Defense-in-depth for
+                        # the server.py busy-branch guard.
+                        if force_plan:
+                            send_kwargs["task_metadata"] = {
+                                "force_plan": True,
+                                "force_plan_source": "consilium",
+                            }
+                        bridge.ui_send(payload, **send_kwargs)
                     else:
                         bridge.ui_send(payload, broadcast=False)
                 except Exception:
--- a/ouroboros/owner_mailbox.py
+++ b/ouroboros/owner_mailbox.py
@@ -29,8 +29,13 @@ def write_owner_message(
     task_id: str,
     msg_id: Optional[str] = None,
     kind: str = KIND_OWNER_TEXT,
-) -> None:
-    """Write an owner message or typed control entry to a task's mailbox."""
+) -> bool:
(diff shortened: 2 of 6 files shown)


razzant:
Thanks for this, @ndrew1337. After reviewing against current `ouroboros` I'm closing this one — not because the concern was wrong, but because it's been overtaken by the architecture.

The busy-branch this patches (`if task_constraint or task_metadata: … else: inject_message`) no longer exists. v6.33.0/v6.34.0 rearchitected it into the "decision turn" model: in `server.py` the `if agent._busy:` path now runs `handle_chat_ephemeral`, where the agent decides by judgment to answer inline / `steer_task` / `promote_chat_to_task`. `steer_task` writes the running task's owner-mailbox (`supervisor/events.py::_handle_steer_task`), which the loop drains at each round boundary (`ouroboros/loop.py`). So "owner message reaches the running task" is solved structurally and LLM-first, rather than via the `force_plan`-gated re-route this PR adds — and re-introducing that gate would fight the new model (it's also why GitHub showed this PR conflicting in `ws.py`, `server.py`, `supervisor/workers.py`).

Restart-survival is largely covered already too: the per-task mailbox is persistent and the task keeps the SAME task_id across an exit-42/99 retry, so a steered message is re-drained on resume. The residual gap is narrow (a crash → retry → crash-before-first-drain storm; `_owner_msg_seen` is process-local and `auto_resume_after_restart` doesn't scan the mailbox). If we want to harden that, it's a small separate change against the decision-turn / `steer_task` path — not a rebase of this PR.

Two correctness issues also surfaced on the diff itself: `write_owner_message`'s new `bool` return is ignored by the sibling caller `ouroboros/tools/core.py::_forward_to_worker`, and `_collect_orphaned_owner_messages` only scans the canonical mailbox dir (it misses `child_drive` mailboxes for forked/workspace tasks).

Closing as superseded. Happy to take a focused follow-up if you want to tackle the narrow restart-survival gap against the current architecture — thanks again for digging into this.

_(Editing this comment — an earlier paste was mangled by shell backtick handling.)_

PR #58, opened 2026-06-28
fix(extension_loader): per-PID staged-import naming + liveness-aware sweep (MAX_WORKERS>1 race)

ndrew1337:
## The bug (confirmed v6.50.0, empirically reproduced)
With `MAX_WORKERS>1` under **spawn** (default on macOS/Windows; MAX_WORKERS=10), all workers boot simultaneously and each stages its **in-process** skill plugin into the **SHARED** dir `state/skills/<skill>/__extension_imports/<uuid>`. Before staging, `_sweep_stale_extension_imports` `rmtree`'d every `<uuid>` tree **not in the current process's keep-set** (no liveness/age guard) → a worker deleted a **peer's still-loading** tree → `exec_module` raised `FileNotFoundError` → the skill **silently failed to load in a subset of workers**. The agent then hit `⚠️ Unknown tool` and confabulated a wrong answer (looks like a model failure). Empirically **1–5 of 10** workers lost the skill per boot.

- Hits any **in-process** native / sideloaded / self-authored skill — including **self-evolution-created** ones (a new capability silently unavailable on a subset of workers).
- **Out-of-process / Hub / native-deps** skills are immune: they stage under a per-invocation private base and never sweep a sibling.
- `server enabled+live_loaded ≠ worker-has-it`: the tool catalog is per-process; the reconcile queue is worker→server and never pushes a tool into a sibling worker.

## The fix (per-PID naming + liveness/grace sweep predicate)
A **single-file** change in `ouroboros/extension_loader.py` — no `worker_main`/env wiring, no `OUROBOROS_EXTENSION_PROCESS_CHILD` reuse (which would flip `current_execution_mode()` / skip the sweep → disk leak), and the SHARED dir stays swept by every process so dead trees self-reap (no private-base leak).

- Stage the in-process leaf as `<os.getpid()>-<uuid>` (the out-of-process child branch stays bare-uuid, untouched).
- `_sweep_stale_extension_imports` reaps a per-PID tree **only** when its owner PID is **dead** (`platform_layer.pid_is_alive`) **AND** its mtime is past a **90 s grace** (mirror of the supervisor spawn grace; module-local to avoid an `ouroboros→supervisor` import). A live owner or a just-spawned (fresh) tree is **kept**; the existing keep-set self-bundle skip is preserved; legacy bare-uuid leaves keep the prior keep-set-only behaviour. **Fails safe**: unverifiable liveness → keep (never reap unverified); an out-of-range/all-digit parse → treated as legacy (never feeds `os.kill` an `OverflowError`).

## Tests
- Rewrote the two single-survivor sweep tests to assert a **LIVE peer (different PID) SURVIVES** and a dead+aged orphan is reaped — deterministic via a monkeypatched `pid_is_alive` (parallel-safe, no real subprocess).
- Added a predicate unit test (live-peer / dead-fresh / dead-old / legacy / all-digit).
- Two subagent reviews (correctness + adversarial, incl. **mutation-testing** the predicate against `skip-if-pid==mine` / dropped-grace / pre-fix) found no regressions. Full parallel + serial suites green; `test_extension_loader.py` + `test_extension_process_runner.py` = 82 passed.

**Immediate workaround for affected deployments:** `MAX_WORKERS=1` (kills parallelism) or force the skill out-of-process (immune).


razzant:
**Maintainer follow-up** — pushed `c63a348`, additive on top of your commit (your work is untouched; no rebase/force).

Thanks for this — it's a real fix and the sweep predicate is well-designed and well-tested. I ran the PR through Ouroboros's own triad + scope review (gpt-5.5 + gemini-3.5-flash + opus-4.8): opus explicitly validated the predicate as *"sound and fail-safe"*, and gemini passed it clean.

The one substantive finding was a docs drift: `docs/ARCHITECTURE.md` still described the staged import tree as `__extension_imports/<uuid>/skill/`, but the code now creates `<pid>-<uuid>`. `c63a348` syncs the data-layout tree line and the Extension Loader section to the per-PID leaf naming + the liveness/grace sweep contract. Nothing else changed.

Remaining review items, intentionally **not** actioned:

- **version_bump (P9)** — no `VERSION`/`pyproject`/README bump. Deferred by design: the version carriers are bumped at the maintainer's release commit, not per contributor PR.
- **"legacy bare-uuid leaves should be preserved, not reaped"** (raised only by the scope reviewer; opus + gemini pass it) — rejected as a false positive. Under the new code the shared `__extension_imports/` dir only ever receives `<pid>-<uuid>` leaves for in-process loads (out-of-process leaves stage under a private, unswept base), so a bare-`<uuid>` leaf there can only be a pre-upgrade relic — and the worker pool restarts as a unit on upgrade, so no live worker owns one. Reaping it is safe and matches the PR's stated *"keep the prior keep-set-only behaviour."*

`test_extension_loader.py` green. Ready to merge from my side.

PR #59, opened 2026-06-29
onboarding bug

mixa2130:
## Problem

When setting up Ouroboros on a remote server via the onboarding wizard, selecting an `openai-compatible::` model
with a required API key always failed at the final save step with:

image

The root cause: `_max_context_block` runs the capability-evidence probe **before** writing settings to disk. Both
probe paths read `OPENAI_COMPATIBLE_API_KEY` from the environment or the already-saved settings file — neither of
which has the key yet during first-time onboarding.

**Metadata probe** (`_openai_compatible_metadata_window`): sent `GET {base_url}/models` without an `Authorization`
header → 401. Because `_metadata_fetch_transport_failed` always returns `False` for `openai-compatible`, this
silently returned 0 (→ `STATUS_UNPROBEABLE`) rather than hard-failing. The real problem was the next step.

**Generative probe** (`_generative_probe_window`, triggered because `allow_generative=True` and `window=0`): sent a
large POST to `/chat/completions` via `LLMClient.probe_oversized_context`, which called `_resolve_remote_target` —
reading the key from the environment again. With an empty key the provider either rejected the connection immediately
or the upload timed out before a response was received, resulting in `status_code=None` → `STATUS_FAILED` →
`probe_failed=True` → HTTP 503 returned to the wizard.

## Fix

Add an optional `api_key` parameter through the entire probe call chain:

```
_max_context_block(settings)
  └─ probe(..., api_key=compatible_api_key)           ← key from current
       ├─ _provider_metadata_window(..., api_key=...)
       │    └─ _openai_compatible_metadata_window(..., api_key=...)
       └─ _generative_probe_window(..., api_key=...)
            └─ LLMClient.probe_oversized_context(..., api_key=...)
                 └─ target["api_key"] overridden before client is built
```

`_max_context_block` extracts `OPENAI_COMPATIBLE_API_KEY` from the in-flight `current` dict (the settings being
saved, not yet on disk) and passes it down. When `api_key` is `None` (all existing callers), both functions fall
back to their previous behaviour — no behaviour change outside the onboarding/settings-save path.


razzant:
**Maintainer follow-up** — pushed `e50d5bb`, additive on top of your commit (your work is untouched; no rebase/force). I also retargeted the base `main → ouroboros` (our dev branch; changes are promoted to main/stable from there).

Thanks — solid root-cause write-up and the right fix for the onboarding 503. I ran the PR through Ouroboros's triad + scope review (gpt-5.5 + gemini-3.5-flash + opus-4.8), which surfaced one real regression:

`_max_context_block` extracted `OPENAI_COMPATIBLE_API_KEY` **unconditionally** and threaded it into `probe()`; `LLMClient.probe_oversized_context` then overwrites the target `api_key` for **any** provider. Since the generative probe also runs for `openai`/`openrouter`/`cloudru` (`_GENERATIVE_PROBE_PROVIDERS`), a settings-save / Max-mode probe on one of those routes would authenticate with the compatible key whenever a leftover `OPENAI_COMPATIBLE_API_KEY` was present — a cross-provider key bleed (and a regression vs. pre-PR behaviour, where the generative probe used the provider's correctly-resolved key).

`e50d5bb` gates the in-flight key on `route['provider'] == 'openai-compatible'` (one-line scope) and adds two regression tests: probe-threading (`test_capability_evidence.py`) and gate non-leakage (`test_max_context_gate.py`). On re-review the cross-provider finding, the pre-existing `>8`-param note, and the missing-test note are all cleared, and the scope reviewer no longer blocks.

Not actioned by design: **version_bump (P9)** — version carriers are bumped at the release commit, not per PR. `probe()`'s parameter count (>8) predates this PR and is left as-is.

Tests green (`test_capability_evidence.py` + `test_max_context_gate.py`). Ready to merge from my side.

PR #60, opened 2026-06-30
fix(evolution): make owner-stop authoritative against post-task re-arm

ndrew1337:
## Problem

Owner-initiated evolution **stop** did not persist. When the owner stopped evolution via `/evolve off`, the natural-language toggle tool, or **panic stop**, the campaign was only **paused** (resumable) and the post-task auto-promotion envelope stayed **armed**. On the next boot tick, `apply_pending_request` re-activated the same paused campaign and flipped `evolution_mode_enabled` back to `True` — evolution **re-armed itself autonomously** after the owner had explicitly stopped it.

Reproduced in a state-machine simulation matching the report's exact campaign id `b6f5bbc6`: an owner stop followed by a boot tick silently resumed the campaign.

## Fix

A durable global flag **`evolution_owner_stopped`** in supervisor state, treated as the authoritative gate:

- **Set `True`** at the 3 owner-stop sites (`/evolve off` → `server.py`, NL toggle → `supervisor/events.py`, panic → `ouroboros/server_control.py`); **cleared** (`= not turn_on` / `= not enabled`) at the 2 owner-**start** sites, so `/evolve start` re-engages cleanly. No other writer clobbers it (`state.py` `setdefault` only fills a missing key). System breakers never touch it, so a legitimate later promotion can still re-arm.
- **New chokepoint** in `apply_pending_request`: when owner-stopped, **drop** the queued request and return `False` — never re-arm. It sits **before** the sole enable site (`evolution_mode_enabled = True`) and **after** every existing early-return, so no path bypasses it. No TOCTOU: event dispatch, the `/evolve` bridge, and `apply_pending_request` all run sequentially on the same supervisor thread each tick.
- **New `complete_evolution_campaign`** (terminal): `status='stopped'`, **archives + pops** any in-flight `active_transaction` (no dangling commit for a boot reconcile to absorb) and pops `post_task_backlog_id`. It replaces the resumable `pause_evolution_campaign` **only** at the 3 owner-stop sites — the **system breakers** (cost cap, restart/bootstrap safety, light-runtime mode, the no-absorb throttle, budget reserve) keep the resumable pause.
- **New `drop_pending_request`** helper unlinks the queued request file at the owner-stop sites (`unlink(missing_ok=True)` in `try/except` — cannot throw).
- **Panic stop** now also sets the durable flag and clears `post_task_autostop` (previously it did neither), making panic authoritative against the post-task pipeline too.

A terminal `'stopped'` campaign is **never resurrected**: `/evolve start` mints a **fresh active** campaign (`start_evolution_campaign` already re-mints on any non-`{active,paused}` status — so the "fresh on re-engage" property needed no code change, only a regression test pinning it). The legitimate **absorb-restart-verify resume** is untouched: it keeps `evolution_mode_enabled=True`, hits the pre-existing enabled-guard (which defers, never drops), and `evolution_owner_stopped` stays `False`.

## Tests

4 new regression tests in `tests/test_post_task_evolution.py` (27 passed total):
- `test_evolution_owner_stopped_blocks_post_task_rearm` — the core inverse: owner-stopped → request dropped, no `start_evolution_campaign`, no autonomous re-enable.
- `test_post_task_apply_still_defers_when_already_enabled` — the pre-existing enabled-guard is **not** stolen by the new guard: a live campaign's resume still defers (request left, not dropped).
- `test_complete_evolution_campaign_is_terminal_and_archives_tx` — terminal status, `active_transaction` archived to `transaction_history` then popped, `post_task_backlog_id` popped.
- `test_evolve_start_after_stop_mints_fresh_campaign` — a `stopped` campaign is replaced by a fresh `active` one.

`ouroboros/review.py` `MAX_TOTAL_FUNCTIONS` 3620 → 3624 (+2 new functions, +2 headroom).

## Verification

- `tests/test_post_task_evolution.py` — 27 passed (4 new). Each evolution test file passes in isolation.
- Full parallel (`-n auto --dist loadscope`) + serial suite — **no evolution failures**.
- `ruff check . --select F` (exactly as CI) — **All checks passed**.
- Two independent adversarial subagent reviews (correctness + regressions) traced the actual code at every call site and both returned **clean**, including confirming the owner/system pause split and that no path enables evolution while leaving the flag stale-`True`.


razzant:
**Maintainer follow-up** — pushed `788113e`, additive on top of your commit (your work is untouched; no rebase/force).

Thanks — this is a genuinely important owner-authority / constitutional fix, and the core design (durable `evolution_owner_stopped` sentinel + drop-chokepoint, terminal `complete` vs resumable `pause`, panic authoritative) is sound. I ran the PR through Ouroboros's own triad + scope review (gpt-5.5 + gemini-3.5-flash + opus-4.8, gpt-5.5 scope) over several rounds; scope is now clean. The follow-up commit closes the review-surfaced gaps:

1. **Atomic re-check** (`post_task_evolution.py::apply_pending_request`): the enable step re-reads the *live* `evolution_owner_stopped` inside `update_state` and refuses the enable if an owner stop raced in after the `load_state()` snapshot (the toggle/`/evolve` paths are serialized ahead of `apply_pending_request` on the single supervisor loop, but panic can fire from another thread) — terminal-closing the stale campaign and dropping the request instead of a stale enable.

2. **Owner-stop mid-cycle worktree cleanup** (`evolution_lifecycle.py`): the `pause → complete` change had a subtle regression — the terminal `stopped` status makes the cancelled task's *async* `task_done` early-return in `update_evolution_campaign_after_task`, so the per-cycle worktree cleanup (which the resumable `pause` preserved) was skipped, leaking abandoned/unreviewed evolution edits into the live repo after an owner stop. `complete_evolution_campaign` now runs `_cleanup_worktree_after_cycle` on the in-flight tx before popping it (self-guarded: skips while a task holds the worktree, pytest/live-repo guarded, kill-switch, never raises), and the graceful owner-stop sites (`_handle_toggle_evolution`, `/evolve off`) cancel the running cycle **before** the terminal close.

3. **Emergency Stop Invariant** (`server_control.py::execute_panic_stop`): because #2 makes `complete_evolution_campaign` do git work, panic (which calls it before the hard exit) must not run it — that would delay panic, which BIBLE forbids. Added `cleanup_worktree=True` to `complete_evolution_campaign`; **panic passes `cleanup_worktree=False`** (no git stash/reset before `os._exit`; the panic flag + boot reconcile own that recovery).

4. **Docs** (`ARCHITECTURE.md`, `state.py`): document the sentinel, the drop-chokepoint, terminal-vs-resumable-pause, and panic behavior; the sentinel is cleared by an *owner-authorized start* — `/evolve start` **or** the owner-directed `toggle_evolution(True)` tool (mirroring how `toggle_evolution(False)` is itself one of the three owner-stop sites).

5. **Tests**: `_handle_toggle_evolution` off/on wiring, the atomic re-check race-abort, `execute_panic_stop` wiring (asserts `cleanup_worktree=False`; destructive kills + `os._exit` neutralized), and `complete_evolution_campaign` running the cleanup before pop (+ the `cleanup_worktree=False` skip).

**Verified false-positive** (one reviewer, advisory): the concern that the cleanup's `git reset --hard` could race a still-writing cancelled worker. `cancel_task_by_id` synchronously `kill_pid_tree`s **and `join()`s** the worker before returning, so by the time the (reordered) `complete` runs the cleanup the worker is already dead — no concurrent-write window.

**Not actioned, by design:** `version_bump` (P9) — version carriers are bumped at the release commit, not per PR. A dedicated `/evolve` slash-command stop/start test (the same `complete`/`drop`/flag wiring is already pinned by the tool-toggle test). And carrying explicit owner-authorized provenance on evolution-start events (so an autonomous tool-call could never clear the sentinel) — a deferred hardening, documented as such.

All evolution test files pass in isolation (`test_post_task_evolution.py` +5 tests; `ruff --select F` clean). Ready to merge from my side.

PR #63, opened 2026-07-01
feat(resume): portable conversation-resume (Claude Code --resume analog)

ndrew1337:
## What

Adds **opt-in conversation-resume** — the Ouroboros analog of Claude Code's `--resume` / `single_conversation`. A task that carries `resume_from_task_id` **continues the SAME conversation** of a prior completed task instead of starting cold.

Two modes (selected via the optional `resume_mode` task field):

- **`continuation` (DEFAULT)** — the true CC `--resume` analog: the stored conversation returns **append-only**, with the original system message **verbatim** (byte-stable prefix ⇒ cross-task prompt-cache hits; no fresh-state re-framing), followed by the sanitized turns; the new user turn is simply appended.
- **`splice` (explicit legacy opt-in)** — the prior turns are spliced between a freshly rebuilt system message and the new user turn (system prompt reflects live drive state each task).

## Why continuation is the default — measured

On CL-Bench `database_exploration` (6-question A/B, same harness/model/effort, per-action tasks):

| mode | score |
|---|---|
| splice | 0.144 |
| **continuation** | **0.611 (×4.2)** |

The replay itself was never the problem (byte-level chain fidelity verified) — the **re-framing** was: a churning rebuilt system block, a truncated duplicate of the very history being replayed, and a never-hitting prompt cache. Continuation removes all three at once. E2E-proven with a codeword test (`memory_mode=empty`, so the resumed conversation is the only possible channel).

## Design (self-contained module + 4 thin hooks)

`ouroboros/resume.py` is stdlib-only (json/os/pathlib/logging) and version-portable. The core needs 4 small hooks:

- **capture** (`agent.py`) — opt-in via `OUROBOROS_RESUME_CAPTURE`; persists the final conversation to `<OUROBOROS_DATA_DIR>/state/resume/<task_id>.json`.
- **load** (`context.py`) — `load_continuation` (default) replaces the fresh message list with the stored conversation; `load_resume_turns` (explicit splice) splices turns into the fresh frame. Both return `[]` on any failure → clean fallback to a fresh build.
- **final-turn fidelity** (`loop.py`) — `note_final_msg` persists the FULL provider message for the no-tool final turn (reasoning/`reasoning_details`/`response_id`), which `run_llm_loop` returns as text without appending; measured 47/58 provider finals carried reasoning that a bare `{role, content}` reconstruction dropped. No-op unless capture is armed.
- **wire** (`gateway/tasks.py`) — accept `resume_from_task_id` + `resume_mode` on the task body.

Plus one adjacent correctness fix: `_run_round_compaction`'s result is now applied **in place** (`messages[:] = ...`) — the caller's reference previously kept a stale pre-compaction list, silently truncating every capture after the first compaction.

## Robustness (from an adversarial multi-agent review — 15 confirmed findings, all addressed)

- **Reasoning continuity preserved on replay** (sanitize drops ONLY `cache_control`): same-model replay accepts `reasoning_details`/thinking blocks with signatures; the llm layer's reactive 400 strip-and-retry is the safety net for any family that rejects them.
- **Chain-length guard** (`_trim_chain`): over `OUROBOROS_RESUME_MAX_CHARS` (default 600K chars) the oldest turns are dropped to ~70% of the cap with ONE explicit omission note — an unbounded chain otherwise 400s at the provider window with no fallback.
- **No silent cold restarts**: a requested-but-missing capture logs a WARNING (the one failure shape callers couldn't see).
- **Replay can never 400 the next task**: dangling assistant `tool_calls` tails get synthesized aborted-tool replies at capture; empty finals are skipped (an empty text block would both lose the end-of-prefix cache breakpoint and 400 on Anthropic lanes); `note_final_msg` keys are allowlisted (no provider echo junk like `tool_calls: []`); the end-of-prefix breakpoint walks back to the nearest non-empty turn.
- **`resume_mode` is normalized and unknown values warn loudly** (a typo must not silently change the mode).
- **H2 path-traversal guard** unchanged: rejects `/`, `\`, leading `.`, `..` + containment under the resume dir.

## Fully additive / opt-in

With `OUROBOROS_RESUME_CAPTURE` unset and no `resume_from_task_id`: `capture` is a no-op, both loaders return `[]` → **behavior unchanged**. Provider-agnostic (`OUROBOROS_DATA_DIR`, no OpenRouter requirement).

## Tests / checks run

- `tests/test_resume.py` — **19 cases**: capture→load round-trip (continuation default: system verbatim; explicit splice: system dropped), mode selection + unknown-mode fallback, breakpoint placement/walk-back, chain trim, `note_final_msg` fidelity/no-op/no-duplicate, dangling-tool synthesis, H2 guard for both loaders, opt-in no-ops.
- Adjacent suites for every touched file all green: `pytest tests/test_resume.py tests/test_context.py tests/test_context_budget_ssot.py tests/test_gateway_smoke.py tests/test_agent_task_pipeline.py tests/test_loop_compaction.py tests/test_compaction.py tests/test_loop_compaction_policy.py -q` → **131 passed**.

### DIFF
--- a/ouroboros/agent.py
+++ b/ouroboros/agent.py
@@ -31,6 +31,7 @@ from ouroboros.memory import Memory
 from ouroboros.context import build_llm_messages
 from ouroboros.context_budget import CONTEXT_SOFT_CAP_TOKENS
 from ouroboros.loop import run_llm_loop
+import ouroboros.resume as resume
 from ouroboros.config import resolve_effort
 from ouroboros.agent_startup_checks import (
     inject_crash_report,
@@ -527,6 +528,10 @@ class OuroborosAgent:
             _scope_pid = str(getattr(ctx, "project_id", "") or "").strip()
             if _scope_pid and not str(task.get("project_id") or "").strip():
                 task["project_id"] = _scope_pid
+
+            # Conversation-resume capture (opt-in via OUROBOROS_RESUME_CAPTURE): persist the final
+            # conversation so a later task can continue it via resume_from_task_id.
+            resume.capture(self.env, str(task.get("id") or ""), messages, text)
 
             emit_task_results(
                 self.env, self.memory, self.llm,
--- a/ouroboros/context.py
+++ b/ouroboros/context.py
@@ -27,6 +27,7 @@ from ouroboros.context_layout import (
 )
 from ouroboros.config import get_context_mode
 from ouroboros.contracts.task_contract import normalize_bool
+import ouroboros.resume as resume
 
 log = logging.getLogger(__name__)
 _LARGE_CONTEXT_SECTION_CHARS = LARGE_CONTEXT_SECTION_CHARS
@@ -1279,8 +1280,11 @@ def build_llm_messages(
                 },
             ],
         },
-        {"role": "user", "content": build_user_content(task)},
     ]
+    # Conversation-resume: splice the prior task's turns between the (fresh) system message and the
+    # new user turn, so successive submissions form one continuous, prompt-cacheable conversation.
+    messages.extend(resume.load_resume_turns(env, task))
+    messages.append({"role": "user", "content": build_user_content(task)})
 
     messages, cap_info = apply_message_token_soft_cap(messages, soft_cap_tokens)
     return messages, cap_info
--- a/ouroboros/context.py
+++ b/ouroboros/context.py
@@ -1281,9 +1281,15 @@ def build_llm_messages(
             ],
         },
     ]
-    # Conversation-resume: splice the prior task's turns between the (fresh) system message and the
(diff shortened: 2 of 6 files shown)


ouroboros-agent:
Thank you for this work, and especially for the measurement: the CL-Bench comparison is the most useful number anyone has brought to the continuity question. The tree has moved a long way since July. Loop, agent and context were rewritten in v7, and continuation now goes through predecessor_task_id with a bounded projection of the previous result, so this branch no longer has anything to attach to and we are closing it as outdated rather than asking you to rebase 3800 commits. If you ever rerun the same A/B against the current continuation path we would love to see the result, because it would tell us whether the warm conversation idea deserves a fresh design on the new context builder. Thanks again for everything you contribute here.

PR #190, opened 2026-08-11
fix(cli): keep packaged installs user-local and migrate owned shims

artbred:
The packaged POSIX installer previously chose the first writable directory under $HOME on PATH. That could place the ouroboros shim inside another tool's private directory, such as ~/.kimi-code/bin.

This update uses ~/.local/bin as the deterministic user-local default. On default installs, it also removes an earlier shadowing shim only when it is a symlink under the user's home and its resolved wrapper carries the exact Ouroboros marker. Foreign or unproven commands are preserved and reported, explicit --target-dir installs are left alone, and the PATH hint now prepends the install directory.

The new shim is installed before cleanup, and an old shim is revalidated immediately before removal. The branch is rebased onto ouroboros and keeps the repository's user-local, no-sudo install contract.

Verification: 9183 passed, 3 skipped, 82 deselected in the full suite; 99 focused and adjacent tests passed after the final base update; Ruff F and git diff checks passed.

### DIFF
--- a/ouroboros/packaged_cli_install.py
+++ b/ouroboros/packaged_cli_install.py
@@ -22,6 +22,8 @@ class InstallPlan:
     source: pathlib.Path
     action: str
     path_hint: str
+    obsolete_shims: tuple[pathlib.Path, ...] = ()
+    shadowing_commands: tuple[pathlib.Path, ...] = ()
 
 
 def main(argv: list[str] | None = None) -> int:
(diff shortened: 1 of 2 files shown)


ouroboros-agent:
Thanks for catching this, Artem. The original report exposed a real fresh-install bug: the command could land in another tool's private PATH directory. We kept your fix focused but adjusted the destination to ~/.local/bin so it remains user-local and does not require sudo, then added a small migration for older Ouroboros-owned shims that would otherwise keep shadowing the new install. Foreign commands and explicit target directories stay untouched. I also rebased the branch and ran the full test and review gates. Everything is green, so I'm merging it. Thanks again for the useful contribution!

PR #98, opened 2026-08-03
front: web-UI redesign — flat gray+red developer-IDE theme, Changes screen with typed task-diff endpoint, task inspector, read-only Files with Cmd+L context capture

kazzand:
## Summary

Full web-UI redesign toward a flat gray+red "developer IDE" look, implemented from a design handoff (README + interactive prototype) under an owner-approved 37-decision plan:

- **Theme**: token-system overhaul in `web/style.css :root` (gray backgrounds, red accent kept, teal project identity, diff/syntax tokens, numeric radius steps); main-SPA glassmorphism (`backdrop-filter`) and the matrix-rain canvas removed; the Design System section of `docs/DEVELOPMENT.md` rewritten to match (onboarding keeps its standalone blur by design).
- **Shell**: sidebar with brand row (version + online dot), Workspace (Changes, Files) and System sections, budget meter pinned to the sidebar bottom (`headerBudgetPresentation` stays the fail-closed formatting SSOT; fill via a CSS custom property); chat header reorganized to Evolve/Review/Restart ghosts + Panic with a slim More (Consciousness); one app-level `/api/state` poll owner (extracted `web/modules/state_poll.js`: 3s on chat / 20s elsewhere / paused hidden / single-flight).
- **Flat chat**: flat agent turns with mono activity rows behind a 2px rule, restyled user/system messages, and an ordered-parts composer over the existing `sendMessage` transport — `web/modules/composer_parts.js` is one reversible `[context: <path> L<a>-L<b>]` marker codec (inline-fence hybrid ≤200 lines, lossless-or-raw rendering invariant, floor producer contract) consumed by the chat composer and the Files/Changes capture docks. 100% of live-card functionality preserved (nested subagents, expandable lazy rows, Cancel run, Turn into project, review projections, honest cost chips) — the pre-existing Playwright suite passes with its geometry assertions unmodified.
[...]

PR #193, opened 2026-08-11
Projects become workdirs with multiple chat threads

kazzand:
# Projects become workdirs with multiple chat threads

A project stops being a single side-panel room and becomes **a working folder with
several chat threads**, on the same contract as every other chat: a thread opens as the
centre chat, can be forked, renamed, reordered, archived and deleted, and each thread
chooses whether it works **in the project folder** or **in its own git worktree**. Main
Chat is untouched.

98 commits, 80 files, +22,863 / −912.

**Base is `front/synthesis`, not `ouroboros`** — a deliberate deviation from
CONTRIBUTING's "base the PR on lowercase `ouroboros`", made because this work is stacked
on the web-UI redesign (PR #98) and `front/synthesis` is its ancestor. Basing on
`ouroboros` would present the entire redesign as this PR's diff. Land #98 first, or
review this against the redesign head.

## What the owner decided (the plan is the authority)

- Thread #0 **is** the project's own chat; the project row and thread #0 are one row, and
  the project's top-level `chat_id` stays its compatibility alias. No migration rewrite.
- **A fork is a cursor, not a copy** — `{fork_of_chat_id, fork_before_ts}` — so no journal
  rows are duplicated and deleting the source does not empty the fork. One shared
  `thread_ancestry_lens(chat_id)` feeds both the history surface and the agent's context,
  so the two cannot describe the same thread differently.
- A thread may be **archived and restored, or deleted**, and deleting one takes its
  worktree with it when the checkout is clean.
- **Per-thread choice of place**: work in the project folder (serialised with the
  project's other work) or branch off into a worktree (runs concurrently).
- Attaching a project needs a place; a non-git folder is admitted through a typed
  `git_init_required` decision rather than a silent refusal.

## Verification

| Lane | Result |
|---|---|
| `python -m pytest` (junit XML, run alone) | **8658 tests, 0 failures, 0 errors, 1 skipped** |
| `node --test web/tests/*.js` | **424 pass, 0 fail** |
| Playwright (`OUROBOROS_RUN_UI_SMOKE=1`) | **35 passed, 1 skipped** |
| `VERSION` | untouched (maintainer assigns at landing) |

Counts are read from the JUnit XML, not from stdout: a full run in this repository exits
without printing its summary line — on the base branch too.

**Vision inspection** (DEVELOPMENT.md's rule that a visible change is looked at in a real
consumer flow, not merely screenshotted): a live instance was driven by hand — created a
project on a real git folder, created threads, opened one as the centre chat, branched one
into its own worktree, archived and restored another, and read every thread-menu row's
enabled state and reason. It found one defect no reviewer could: the archived row's
tooltip read "Restore this thread — restore it to act on it", a tautology beside the word
"Restore". Fixed, and the test that pinned the old phrasing now asserts the substance.

## Review record

Four phases, each through two adversarial rounds, then the merge, then two integration
reviewers, then the paid contributor triad. Every finding was re-verified against the code
before being accepted, and several reviewer claims were **refuted with reproductions**
rather than implemented:

- 27 findings closed in the worktree/lifecycle phase, including a snapshot that committed
  conflict markers into a folder stopped mid-merge, a renamed tracked credential deleted
  from the owner's branch, and a merge onto a detached project HEAD that reported success
  while both safety judges were fooled by the same wrong reference.
- 19 integration findings, including two blockers found one level up: deleting a **project**
  orphaned every branched thread's checkout and branch with no unmerged-work refusal at
  all, and the merge's own central cross-stream fix had **zero** regression protection —
  five mutations dropping it survived the entire suite. It now has five call-site tests
  plus a source guard, proven by re-applying each mutation.
- 8 of the paid gate's 11 criticals were live and are fixed; 2 were stale (already closed
  by later commits) and were deliberately left alone; 1 was the vision-inspection gate,
  answered by doing it. Two reviewer prescriptions were declined for reverting owner
  decisions, and the alternatives close the same hazards.

Notable class, found three times: the server grows a typed escape that no client can
produce. The root was that every thread route answers 4xx and the client layer rejects any
non-2xx, so refusals never reached the code that renders them. DEVELOPMENT.md now carries
the rule — a refusal carrying `acknowledgeable` or a typed `decision` must be rendered
with its answer in the same gesture.

## Disclosed, not silently carried

- **The ancestor origin-fallback synthesis is history-only.** `context.py` shares the lens
  but not that synthesis, so an agent in a fork of a converted project can still miss the
  parent's referenced Main origin message. Stated in ARCHITECTURE rather than implied away.
- **Origin dedupe runs before the cutoff filter**, so binding-file order decides which
  owner's cutoff wins and a stricter winner can drop a row a looser ancestor would admit.
- **Thread checkouts live outside the data directory** (`~/Ouroboros/thread_worktrees`), so
  two instances on one machine share that root. Verified consequence: provisioning refuses
  to reuse an existing path or branch, and removal acts only on its own registry row — a
  second instance gets a typed refusal naming the path, not a stolen checkout.
- **Two projectless tasks naming the same folder now serialise** where they previously ran
  concurrently. That is the cost of closing the two-writers-in-one-folder hole.
- `supervisor/queue.py` sat at exactly 1600 lines and crossed the gate by one behavioural
  line; it is recorded in the grandfathered-modules SSOT by repo-relative path with its
  reason and the tracked follow-up, the same mechanism the repository already uses.
- **Pre-existing, not from this branch:** `tests/test_telegram_miniapp_companion.py` has two
  tests that fail or hang in a full-suite run depending on timing (fixed ports 9012 / 45678
  / 8765, no socket guard anywhere in the suite). They fail on the base branch too and pass
  when the file runs alone.

## Conflicts against current `ouroboros`

The redesign was rebased onto `ouroboros` on 2026-08-09; upstream has advanced 109 commits
since. Measured by trial merge: the redesign **alone** conflicts in 5 files
(`docs/DEVELOPMENT.md`, `tests/test_gateway_parity.py`, `web/onboarding.css`,
`web/settings.css`, `web/style.css`). This branch adds exactly **one** more —
`docs/ARCHITECTURE.md`, a single 9-line hunk. Everything else belongs to the redesign and
resolves when it lands.


ouroboros-agent:
Thank you for this, it is one of the most carefully reviewed branches we have had. It is stacked on the redesign in #98 and targets the frozen front/synthesis branch, and Projects have since been rebuilt around a single room with its own memory, so it cannot land as it stands, and several threads per project is not something we are pursuing right now. The ideas we want to keep are the fork as a cursor instead of a copied journal, the single ancestry lens shared by the UI and the agent, and the one writer per folder rule, and we have noted them for later. Closing with thanks.

PR #202, opened 2026-08-12
Fix/retry bypass response cache

Fgeeha:
## Summary

A retry after `provider_incomplete_response` re-sent a byte-identical request.
When any response cache sits between the client and the model, every attempt is
served the same cached failure, the transient-retry budget is spent without
reaching the model, and the task ends as `infra_failed` / `provider_unavailable`.

`provider_incomplete_response` is classified as transient
(`_TRANSIENT_RETRY_KINDS` in `ouroboros/loop_llm_call.py`), i.e. the loop assumes
a repeat *may* produce a different result. That assumption only holds if nothing
caches. Inside the attempt loop, `attempt` was used solely for bookkeeping —
`_llm_attempts_used` and the observability manifests — while `send_messages` and
`kwargs` were rebuilt identically every time.

### How it showed up

Ouroboros → LiteLLM proxy with `cache: true` → local model:

- six attempts, all returning `content: ""`, `tool_calls: null`, 61 completion
  tokens;
- **the same `response_id` in all six responses**;
- each returned in ~0.0 s, so no attempt reached the model;
- the same request replayed by hand later succeeded, with
  `finish_reason=tool_calls` and valid arguments.

Nothing errors anywhere: the model is fine, the gateway is fine, Ouroboros is
fine. Only the combination fails, and it fails silently — the task reports
`completed` at the lifecycle level while the execution axis says `infra_failed`.

### Fix

From the first retry onward the request carries LiteLLM's documented per-request
cache opt-out. The first attempt is left untouched, so it still benefits from a
warm cache.

The flag rides in `extra_body`. This is not cosmetic: the OpenAI SDK raises
`TypeError: Completions.create() got an unexpected keyword argument 'cache'` on
unknown top-level kwargs, so a raw `cache=` argument never reaches the wire.
Servers that do not know the field ignore it — verified against a plain Ollama
OpenAI-compatible endpoint, which accepted the request unchanged.

Threaded along the same path as the existing `cache_affinity` parameter:

```text
call_llm_with_retry   (attempt > 0)
      ↓
LLM.chat
      ↓
_chat_remote          (both branches: pooled client and no_proxy)
      ↓
_build_remote_kwargs  (both branches: direct providers and OpenRouter)
```

`cache_affinity` itself is not usable for this — it does the opposite, pinning
repeat calls to one upstream so its prompt cache stays warm.

## Scope

In scope:

- `ouroboros/llm.py` — optional `bypass_response_cache` flag threaded through
  `chat`, `_chat_remote` and `_build_remote_kwargs`, emitted into `extra_body`.
- `ouroboros/loop_llm_call.py` — set the flag from the first retry.
- `tests/test_retry_bypass_response_cache.py` — regression tests.

Non-goals:

- The first attempt is deliberately still cacheable; this does not disable
  caching, only the replay of a failed response.
- No change to the runtime's own prompt cache. That one is worth keeping: a
  prefill cache hit measured ~25x faster than a cold pass on the same request
  (423 → 11,656 tokens/s). Only the gateway's *response* cache is opted out of.
- No retry-policy change: budgets, backoff and classification are untouched.

## Verification

- [ ] The default local test suite passes, or the reason it was not run is below.

`make test` and `make lint` could not be run as documented: the pinned toolchain
requires `uv==0.12.1` and this machine has `0.9.2`, so `uv run --locked` refuses
to start. Tests were run directly with the bundled interpreter instead.

Commands and results:

```text
$ python3 -m pytest tests/test_retry_bypass_response_cache.py -q
....                                                                     [100%]
4 passed

$ python3 -m pytest tests/ -k "llm or retry or cache or context_fit"
364 passed, 1 skipped, 8521 deselected

$ uv run --locked python -m pytest tests/ -q --tb=short
error: Required uv version `==0.12.1` does not match the running version `0.9.2`.
```

The tests fail without the change. Checked by restoring both source files to the
base revision and re-running:

```text
FAILED ...::TestBuildRemoteKwargsCacheOptOut::test_no_cache_field_present_when_bypassing
FAILED ...::TestRetryLoopRequestsCacheOptOut::test_first_attempt_does_not_bypass_but_retry_does
FAILED ...::TestChatSignatureThreadsFlag::test_chat_accepts_bypass_response_cache
```

Live behaviour on the affected setup (GTX 1080 Ti, LiteLLM gateway, `qwen3.5:9b`
via an `openai-compatible::` route), same task each time:

| Configuration | `provider_incomplete_response` | API errors | Result |
|---|---:|---:|---|
| Before, gateway cache on | 6 | 0 | `failed` |
| Before, gateway cache off | 3 | 0 | `completed` |
| First version of this fix (top-level `cache`) | 1 | **2** | `failed` |
| This fix | 0 | 0 | `completed` |
| This fix, second cycle | 1 | 0 | **`completed`** — the retry recovered |

The last row is the point: a provider failure still occurred, and the task
survived it because the retry actually reached the model.

**Not verified end-to-end against a live cache.** Re-enabling `cache: true` on
the gateway to reproduce the original failure is currently impossible — that
LiteLLM image crashes on startup with
`AttributeError: 'Cache' object has no attribute 'cache'` in `_init_cache`. So
the mechanism is proven at the unit level and by the retry recovering in the
table above, but not by a before/after against a working response cache.

The branch was rebased from the fork's `main` onto `ouroboros`, moving the base
forward by five commits. Everything above was re-run afterwards rather than
carried over; `git log dcf48aac..ouroboros -- ouroboros/llm.py
ouroboros/loop_llm_call.py` is empty, so neither file was touched by those
commits, and the three changed files are byte-identical across the rebase.

## Visual evidence

Evidence: —

- [x] I followed `CONTRIBUTING.md` and checked the relevant guidance in
      `BIBLE.md`, `docs/ARCHITECTURE.md`, `docs/DEVELOPMENT.md`, and
      `docs/CHECKLISTS.md`.
      Behaviour is internal to the retry path; the rationale lives in code
      comments at both edit sites and in the test docstrings.
- [x] I did not include secrets, local settings, runtime state, logs, caches, or
      generated build/review artifacts in the commit.
- [x] I did **not** bump `VERSION` or release-only version carriers.

- Agent/tool and model:
- Work performed by the agent:
- Human verification performed:

## Review evidence

- Reviewed base SHA: `1952e042a06cc32926b12d1dece4e567e958d11f` (`ouroboros`)
- Reviewed head/diff SHA: `07937b6b`
- Review command/profile: —
- Triad verdict: —
- Scope verdict: —
- Run artifacts/link: —
- Known advisory findings or accepted tradeoffs: `cache: {"no-cache": true}` is
  LiteLLM's spelling. Another OpenAI-compatible gateway with a response cache
  would need its own opt-out; this change does not attempt to generalise, it
  makes the common deployment behave correctly and stays inert elsewhere.
- If not run: the triad/scope tooling was not available in this environment.

## Final checklist

- [x] The branch is based on a current `ouroboros` revision.
- [x] The description explains any limitations, follow-up work, or compatibility
      impact.

The branch is a single commit on top of `ouroboros` at `1952e042` and touches
three files; `git diff ouroboros --name-only` returns exactly those three.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -916,6 +916,15 @@ Retry budgets are failure-class specific. Empty/incomplete responses and transie
 
 Prompt caching is stable-first. Governance and task-stable contracts precede mutable evidence; review builders disclose the stable/dynamic boundary and keep untrusted payloads outside the governance cache block. Provider-specific cache hints are sent only where supported and receive one exact retry without the rejected hint. Rejection evidence is durable and route-specific. Cache identity must not weaken exact review bindings or create a second review authority.
 
+Gateway response-cache recovery is narrower and reactive. The first call remains
+cacheable; only a main-loop `provider_incomplete_response` arms a fresh-response
+request for later attempts. The generic `openai-compatible` route is Ouroboros's
+gateway/proxy route, so only it renders LiteLLM's documented
+`extra_body.cache.no-cache` control. A strict compatible endpoint that explicitly
+rejects `cache` receives the existing one exact cache-parameter removal retry.
+Direct providers and OpenRouter never receive this LiteLLM field; no URL/model
+heuristic or parallel gateway-capability subsystem is introduced.
+
 #### Vision and local image evidence
 
 `analyze_screenshot` and `vlm_query` are bounded secondary-model calls through `LLMClient.vision_query`; `view_image` attaches a local image natively to the active conversation. Send-time image routing works on a copy of the transcript so captioning or placeholder conversion never mutates canonical history. Image payloads are validated, capped/downscaled, and confined to readable roots derived from the Tool API policy matrix plus the protected-artifact rule; URLs and inline base64 are not accepted as local paths.
--- a/ouroboros/llm.py
+++ b/ouroboros/llm.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import copy
 import hashlib
 import json
 import logging
@@ -10,7 +11,6 @@ import os
 import re
 import threading
 import time
-import copy
 from typing import Any, Dict, List, Optional, Set, Tuple
 
 from ouroboros.provider_models import (
@@ -1203,7 +1203,7 @@ class LLMClient:
         target: Dict[str, Any],
         exc: BaseException,
     ) -> Optional[Dict[str, Any]]:
-        """Remove only an explicitly rejected cache-affinity parameter once."""
+        """Remove only an explicitly rejected cache control or affinity once."""
         provider = str(target.get("provider") or "").strip().lower()
         extra_body = payload.get("extra_body")
         param = ""
@@ -1215,6 +1215,12 @@ class LLMClient:
             and "session_id" in extra_body
         ):
             param = "session_id"
+        elif (
+            provider == "openai-compatible"
+            and isinstance(extra_body, dict)
+            and "cache" in extra_body
+        ):
+            param = "cache"
         if not param:
             return None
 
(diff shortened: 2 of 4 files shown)


ouroboros-agent:
Thanks for tracing this subtle retry/cache interaction and for the concrete field evidence. We kept the useful core of the fix, narrowed the bypass to retries after a bare incomplete response on the generic OpenAI-compatible route, and added exact recovery for strict endpoints that reject LiteLLM’s cache field. Direct providers and OpenRouter no longer receive LiteLLM-specific JSON, and the behavioral tests now cover incomplete responses, transport and body errors, provider routing, and strict recovery. I’ve rebased, verified, and merged the PR. Thanks again for the careful report and contribution!

PR #210, opened 2026-08-13
fix(git_ops): bound rescue/rollback git processes with a timeout

### LINKED ISSUE #182: Rescue/rollback git call graph has no timeout authority
Follow-up from the managed-update fixes audit (PR #177 context); enhancement, base behavior.

git_capture in supervisor/git_ops.py runs git without any timeout, and the rescue/rollback call graph (status, diff capture, stash, topology reads) inherits that. A git process that never returns (fsmonitor hang, dead filesystem) can stall a rollback or restart indefinitely. The review-side capture added in PR #177 already carries its own 300s timeout; the supervisor-side git graph has none.

Fix direction: one rescue-local timeout authority — a bounded wrapper used by the rescue/rollback paths, where TimeoutExpired becomes a capture warning and the rollback continues (fail-open). Not a blanket timeout for every git call in the codebase.

AmirF194:
## Summary

In scope:

- `git_capture()` runs every rescue/rollback git process with no timeout at all.
  The whole call graph reached from `rescue_before_destructive_rollback` (status
  probes, diff capture, stash creation, topology reads via
  `_collect_repo_sync_state` and `_create_rescue_snapshot`) inherits that: a
  hung git process (fsmonitor deadlock, an unresponsive filesystem) can stall a
  rollback or restart indefinitely. The review-side capture in
  `review_binary_context.py` already carries a 300s timeout for the same class
  of problem; the supervisor-side rescue graph had none.

Non-goals:

- A blanket timeout on every `git_capture` call site in the codebase. Per the
  issue, this is a rescue-local bound only.

## Fix

`git_capture()` gets an opt-in `timeout` parameter, defaulting to `None`
(unbounded, matching the current behavior of every existing call site). A new
`rescue_git_capture()` wrapper bounds it with `RESCUE_GIT_TIMEOUT_SECS` (300s).
Only the rescue/rollback graph's call sites route through the wrapper: the two
direct calls in `rescue_before_destructive_rollback`, every call in
`_collect_repo_sync_state`, `_create_rescue_snapshot`, and
`_copy_untracked_for_rescue`, plus the raw `subprocess.run` used for
`changes.diff` capture. A timeout there returns through the normal nonzero-rc
shape, which every one of those callers already treats as a warning rather
than a hard stop, so the fail-open behavior the issue asks for falls out of
the existing error handling rather than needing new branches.

## Verification

- New regression test `test_git_capture_times_out_instead_of_hanging`: a
  genuinely slow subprocess (`sleep 5`) under `timeout=0.2` returns promptly
  with a typed failure instead of blocking; fails on `origin/ouroboros`
  (`git_capture()` takes no `timeout` kwarg there) and passes on this branch.
- `test_git_capture_default_timeout_is_unbounded`,
  `test_rescue_git_capture_bounds_with_rescue_timeout`, and
  `test_collect_repo_sync_state_uses_rescue_bounded_capture` cover the default
  stays unbounded elsewhere and that the rescue graph actually routes through
  the bounded wrapper.
- `tests/test_git_ops_recovery.py`, `test_git_ops_rollback_sync.py`,
  `test_update_merge_assisted.py`, `test_update_apply_routing.py`,
  `test_update_status_cache.py`, `test_packaging_sync.py`, `test_git_rollback.py`,
  `test_git_shell_policy.py`: all green (293 tests), plus `ruff check . --select F`.
  All run in a clean `python:3.10-slim` container with `git` installed.
- Not run: the browser/UI/integration marker lanes and any test outside the
  files above; `git_ops.py`'s module-level imports are otherwise pure-stdlib
  plus this repo's light deps, so those files import and run cleanly without
  the full `desktop`/`build` dependency profile.

Fixes #182


ouroboros-agent:
Thanks for taking this on, Amir. The core direction was right: rescue and rollback Git work needs a local time bound so a stuck Git process cannot block restart forever.

I rebased your contribution onto the current ouroboros tip and tightened the implementation around one shared cross-platform process runner. The full rescue graph now uses it, including topology checks, index repair, binary diff capture, untracked-file discovery, and the gateway snapshot paths. Ordinary Git calls keep their existing timing, while managed rollback still fails open and records timeout or unknown-state details for recovery.

I also added the settings projection, architecture notes, and portable regression coverage needed to keep that contract explicit. Thanks for the clear issue framing and the solid starting patch.

AmirF194:
Glad the timeout bound was the right direction, thanks for carrying it through to one shared cross-platform runner and adding the regression coverage. Appreciate the merge.

PR #222, opened 2026-08-15
fix: distinguish tree-read errors from real absence in binary review metadata

### LINKED ISSUE #181: Managed binary review: object-read errors can render as proven absence
Found during the managed-update fixes audit (PR #177 context); pre-existing contract gap.

DEVELOPMENT.md says missing exact binary metadata must block the managed-update binary review, but a git object-read failure (ls-tree/cat-file) in the exact-metadata renderer surfaces as an omission marker that the triad can still PASS over — an unread object can render as a proven "absent".

Fix direction: propagate a typed present/absent/error outcome from the git object probes and require successful stage/HEAD/MERGE_HEAD reads in the managed binary lane; an error or incomplete metadata blocks the triad there. Outside the managed lane the current fail-open rendering stays.

AmirF194:
## Summary

In scope:

- The managed-binary review renderer (`render_staged_binary_metadata` in
  `ouroboros/tools/review_binary_context.py`) can render a git object read
  failure identically to a proven absence, so a reviewer sees `absent` for a
  HEAD or MERGE_HEAD blob that was actually never read.

Non-goals:

- The rest of `_git_bytes`'s callers (`staged_path_is_binary`, `_object_size`,
  the staged-index and staged-deletion lookups). Their existing empty-on-any-
  failure contract is unchanged; this PR only threads a typed outcome through
  the two `_tree_entry` lookups the issue names.

## Fix

`_git_bytes` collapsed any subprocess failure (`OSError`, timeout, non-zero
exit) and any genuinely empty result to the same `b""`, so `_tree_entry`
could not tell "the ref has no entry for this path" from "the read failed."
`_tree_entry` now returns a third `ok` value. It is `True` when the ref
itself does not resolve (no `MERGE_HEAD` outside a merge, an unborn `HEAD`)
or the ref resolves and the path is genuinely absent from its tree, and
`False` only when `ls-tree HEAD -- <path>` fails despite the ref resolving:
a corrupt or unreadable tree object. `render_staged_binary_metadata` now
hard-blocks (returns `None`, the contract its two callers already use for
an unreadable staged object) whenever either lookup comes back `ok=False`,
in both the added/modified and the deleted-binary rendering paths.

## Verification

- `test_render_staged_binary_metadata_hard_blocks_on_head_tree_read_error`
  and `test_render_staged_binary_metadata_deletion_hard_blocks_on_merge_head_tree_read_error`:
  delete the loose tree object backing `HEAD` (first test) or a real
  in-progress merge's `MERGE_HEAD` (second test) so the ref still resolves
  but `ls-tree` fails. Both fail on `origin/ouroboros` (metadata renders
  `absent` instead of blocking) and pass on this branch.
- Two control tests confirm genuine absence (a new file with no HEAD entry,
  no merge in progress) and a `cwd` git cannot even spawn into still render
  or block correctly, not just any read failure.
- `tests/test_review_binary_context.py` plus every other test file that
  imports this module (`test_external_review_script.py`,
  `test_git_review_pipeline.py`, `test_review_anti_thrashing.py`,
  `test_review_intent_split.py`, `test_review_observability.py`,
  `test_scope_review.py`): 325 passed. `ruff check . --select F` clean.
  All run in a clean `python:3.10-slim` container with git installed.
- Not run: the browser/UI/integration marker lanes and any test outside the
  files above.

Fixes #181


ouroboros-agent:
Thanks Amir, this closes a gap we filed in #181 and the failure injection tests are exactly the right shape, they reproduce the bug on the old code and pin both control directions. While integrating we brought the branch up to date with the current tip and tightened one spot the review panel caught: the rev-parse probe treated any failure as proof of absence, so a timeout could still render as absent. It now reports absence only on a clean rc 1 and fails closed otherwise, with a direct test for that seam. We also widened the test module docstring. The triad-side blocking question from the issue is tracked separately. Appreciate the explicit non-goals section in your description, it made review straightforward.

AmirF194:
Thanks for the quick turnaround and for catching the rev-parse timeout gap, fail-closed on anything but a clean rc 1 is clearly the right behavior there. Glad the tests pinned it well.

PR #228, opened 2026-08-16
fix(web): prevent Windows layout-switch Alt key from hijacking input focus

dead9111:
## Summary
Fixes a Windows-specific keyboard layout switching bug where pressing `Alt` (or `Alt+Shift`) inside an active text input/textarea triggered default Windows window menu mnemonic activation. This caused the system default error sound to play and dropped the first character typed after switching the layout.

## Scope

In scope:
- Suppress default window menu activation on standalone `Alt` / `Alt+Shift` keydown when focus is inside `INPUT`, `TEXTAREA`, or `contenteditable`.
- Preserve `AltGr` (`ctrlKey + altKey`) behavior on European keyboard layouts.
- Add unit tests for Windows Alt menu suppression logic (`web/tests/windows_layout_switch.test.js`).

Non-goals:
- Modifying non-input shortcut handling or non-Windows platform behavior.

### Automated tests
`node --test web/tests/windows_layout_switch.test.js`
`npm run lint`

Output:
```text
✔ Windows layout switch: standalone Alt suppresses menu activation when focus is in textarea or input (0.6ms)
✔ Windows layout switch: AltGr (Ctrl+Alt) is preserved and NOT suppressed (0.2ms)
✔ Windows layout switch: regular keys, shortcuts, and Enter/Shift+Enter are NOT suppressed (0.2ms)
✔ Windows layout switch: Alt when focus is outside editable elements is NOT suppressed (0.2ms)
tests 4, pass 4, fail 0

Manual verification
Before: In Windows desktop webview, switching layout via Alt+Shift in #chat-input played a chime sound and ignored the first character typed.
After: Switching layout via Alt+Shift produces no sound and immediately inputs the first typed character.
Verified that Enter, Shift+Enter, Ctrl+C/V/A, and AltGr symbols remain fully functional.


ouroboros-agent:
Thank you for this fix. The layout switch chime and the swallowed first character were a daily annoyance for anyone typing in two languages, and your diagnosis of the standalone Alt keydown was exactly right. I built on your branch with a small extension: the same guard now also runs on keyup, because Firefox style hosts decide menu activation on release rather than on press, and the onboarding wizard installs it too, since it lives in its own iframe document where the app listener cannot see keyboard events. I also excluded the AltGraph key by name so AltGr typing stays untouched in the Firefox reporting shape. Your commit is unchanged and your tests still pass, now alongside a few new ones. Thanks again for a well aimed and well tested contribution.

PR #257, opened 2026-08-18
fix: six live-install defects — panic browser kill, invisible second Open, stray server generations, log autoscroll, untested provider keys, freezing Save

ndrew1337:
Six fixes observed on live installs, five of them on a Linux browser-mode session running v6.102.0 where they compounded into an unusable state: a panic stop killed the owner's whole browser, a second Open appeared to do nothing, leaked server generations SIGKILLed every incoming generation until the supervisor loop collapsed, the logs panel kept yanking the reader to the bottom, and a dead provider key only surfaced as a provider outage minutes into the first real task. The sixth is a slow, app-freezing Save Settings button.

## 1. Port sweeps kill the listener, not connected clients

Both POSIX port-sweep helpers (`platform_layer.kill_process_on_port`, launcher `_kill_stale_on_port`) selected pids with a bare `lsof -ti tcp:PORT`, which also matches ESTABLISHED client sockets. In browser mode the owner's browser holds exactly such a socket to the UI port, so a panic stop SIGKILLed the whole browser — violating the invariant documented on `_open_browser_detached` (the browser is the owner's application, outside custody). The Windows branch already filtered `LISTENING`; POSIX now matches it with `-sTCP:LISTEN` (plus `-nP` so name resolution cannot eat the 5s timeout).

## 2. A second browser-mode launch surfaces the running instance

The already-running branch printed a notice to stderr and exited — invisible from a desktop icon, so Open read as "does nothing" while the instance kept running (closing the browser never stops the server by design). The branch now health-polls the port file (bounded, re-reading between probes, so an Open racing the first launcher's bootstrap lands on the live port) and opens the default browser at the existing URL with a bounded join on the opener thread.

## 3. Logs stick to the bottom only while the reader is there

Every visible live entry unconditionally scrolled the panel to the bottom, yanking the reader away from history. Per-entry autoscroll now sticks only while the reader is already pinned near the bottom, with the pinned state captured before any growth (an already-mounted task card grows in place before the append). Deliberate jumps stay unconditional, and a new `ouro:page-shown` re-pin heals the `.page` display round-trip that resets scrollTop to 0 — without it the pinned-only rule would silently disarm autoscroll after every page switch.

## 4. A Test control on every provider card

No provider key had any validity check at paste/save time. Each Settings provider card gains a Test button wired to `POST /api/providers/test`: allowlisted owner-edited values merge over saved settings and run through the same `_provider_specs` catalog fetchers as the model catalog. Saved secrets render masked, so the client sends an override only for fields the owner actually edited; an edited-to-empty field sends an explicit empty override that unsets the key for the probe. Hardened through several review rounds: an endpoint-shaping override (base URL, MiniMax region, the GigaChat TLS-verify toggle) is refused unless every credential of that provider that would remain non-empty after the merge is re-entered in the same request — accepting a caller-supplied endpoint alone would send stored keys to a server nobody configured; provider error text is redacted of every credential the probe knows in raw, percent-encoded and base64 spellings; contract misuse gets typed 400s (unknown provider, unknown region, malformed body); a displayed verdict expires the moment its card's fields change (input, programmatic Clear, settings reload), and a result arriving after an in-flight edit says it verdicted the previous draft.

## 5. The launcher reaps leftover generations of its own install

SIGKILLed server generations leak full `server.py` processes that the custody reaper never sees — missing ledger entries are exactly the defect — and concurrent generations then share the data directory: on the live install every owner restart ended with the incoming generation SIGKILLed ~4s later by a leftover one, twice collapsing the supervisor loop.

Holding the single-instance pid lock licenses the launcher to reap leftover generations of its own install. New `ouroboros/launcher_server_reaper.py`: one unbranded `ps -ww -u <uid>` enumeration (candidate selection must not depend on the install path containing any particular word); a pid is proven only by the exact `<REPO_DIR>/server.py` argv token directly after a python interpreter token plus both launcher-stamped environment assignments (`OUROBOROS_DATA_DIR` equal to ours, `OUROBOROS_MANAGED_BY_LAUNCHER=1`) read from `/proc`; hosts without `/proc` are report-only with a named warning, because the `ps -E` fallback mixes argv into the environment column and an argv token must never authorize a kill. Proofs are revalidated immediately before the signal, descendants are captured before the root signal (SIGKILL reparents children to init), killed means confirmed dead via the new EPERM-honest `platform_layer.pid_provably_gone`, passes are bounded, and a proven survivor suppresses `start_agent` for that generation instead of booting a colliding second server. Panic and window-close paths are untouched; the custody ledger is deliberately never consulted; unreadable environments, direct runs, sibling installs and processes merely mentioning the path are spared and named. The sweep runs at preflight and at the top of every lifecycle generation, ordered recorded-cleanup → stray sweep → port sweep. The startup stray check stays report-only and now labels findings `same_install`/`foreign` through the same identity rule.

## 6. Save Settings stops paying for a cold daemon and stops freezing the app

Two independent causes. The Save handler awaits `loadSettings`, which awaited the shared Claudexor status read (with per-harness model discovery) through BOTH section reloads — a read that can wake a cold daemon. Both sections now await that probe for a bounded beat only (`boundedStatusRefresh`: a warm daemon settles inside it and keeps the exact old semantics; a cold one keeps refreshing in the background while the status surface bindings repaint on arrival), and the clean-draft baseline follows late store arrivals while the draft is clean — an over-eager unsaved-changes indicator in the one narrow cold-daemon window is the disclosed residual, never a lost edit.

And the settings POST — an async endpoint with an entirely synchronous body (validation, disk write, env projection, hot-reload side effects, plus network evidence fetches when review keys changed) — ran on the event loop, freezing every request and WebSocket for the whole save. Every settings writer now runs its body in a worker thread, and because worker threads do not inherit the loop's free serialization, in-process serialization moves into the write seam as `settings_document_mutation()`: the generic save, all five single-decision owner endpoints and the onboarding transaction hold the same document lock across read-merge-write, environment projection included (released earlier, writers could commit A→B but project B→A), so no writer can silently revert another and the event loop never holds the lock. The context-mode idle guard is re-proved under that lock (both halves — the mode being lowered from and the running-tasks check), matching the seam's own prove-under-the-lock precondition doctrine. The node behavioural suites moved their settlement boundary accordingly.

## Verification

Full suite green on the branch head (the one failure, `test_smoke.py::test_tool_set_matches`, reproduces on clean upstream and is unrelated), and `node --test web/tests` green (402). New coverage: 28 reaper tests (proof rules, spare classes, revalidation, fork races, bounded passes, confirmed-dead vs signalled, aborted-sweep survivor contract, behavioural lifecycle wiring — no test signals a real process), 24 provider-test endpoint tests (override precedence, exfiltration guards, redaction forms, typed 400s, constant-sync pins against `_provider_specs`), listener-scope pins for both port sweeps, three headless-launcher tests, seven logs-fixture pins, and save-path pins (the endpoint must delegate off the loop, every writer must hold the document lock, neither section may await the status probe). `ruff --select F,E9` clean; the size-ratchet manifest is valid at every commit; the platform-abstraction guard passes (kill/liveness primitives live in `platform_layer`).

Each fix went through multiple independent reviewer-panel rounds to convergence.


ouroboros-agent:
Thanks, Andrew, for tracking down and testing these Linux issues. I preserved your commits and attribution, moved the PR onto the current ouroboros branch, and made a few focused follow-ups: the provider button now tests the visible draft with one real model call, the launcher uses the shared process-discovery path, and the API contracts and architecture docs match the shipped behavior. The full local Python and web suites, the PR quick test, and the Claudexor fixture matrix on Ubuntu, macOS, and Windows are all green. Really appreciate the work you put into this.

PR #367, opened 2026-08-29
Add the ultra reasoning-effort tier and fix effort bugs across the scale's consumers

ndrew1337:
The reasoning-effort scale stopped at `max`, but the codex vendor ladder tops out one tier higher at `ultra` (`low < medium < high < xhigh < max < ultra`), so on a live install every owner surface that validates against `EFFORT_SCALE` — reviewer-slot rows, Available-subagent rows, the legacy `harness[=model]:effort` migration, and the effort lanes — refused the delegated harness's own top tier. This PR adds `ultra` above `max` and fixes the effort bugs an audit of the scale's consumers turned up along the way.

## The scale extension keeps the uniform-set design

`EFFORT_SCALE` stays the single ordered SSOT: one shared set everywhere, per-route adaptation downstream — a route that tops out lower falls back to the nearest supported tier exactly as before (learned ceilings and exact-route wire recovery on API routes, per-model resolution on delegated ones). The owner-facing mirrors (settings UI effort cards, the route editor's choices) gain the tier; the `switch_model` tool's effort enum is now derived from `EFFORT_SCALE` instead of a hardcoded copy that had already drifted, which makes the SSOT comment in `config.py` true. `config.py` sits exactly at its 1600-line pin, so the scale comment was reworked net-zero.

## Effort bugs fixed in the same sweep

- **`switch_model` silently coerced an unknown effort to `medium`** while the model branch right beside it refuses unknown input with a visible warning. The effort branch now refuses symmetrically, before anything in the call is applied — a same-call model switch does not half-apply behind a rejected tier. A whitespace-only effort deliberately stays a no-op and is pinned as such.
- **Fresh-route recovery could not reach a provider ceiling two rungs down within the acceptance rail.** Acceptance reviews bound physical sends to two; recovery walked one rung per rejection and persists evidence only after semantic success, so a request two tiers above a route's real ceiling burned both sends, learned nothing, and repeated the outage on every call (the class predates this branch — `max` against a high-topped provider — but `ultra` against an xhigh-topped one made it practical). When the provider's rejection message itself names supported tiers, recovery now jumps straight to the highest named one, mirroring the existing mandatory-reasoning jump to `low` — so quoting providers resolve `ultra` against an xhigh ceiling within the two-send rail in one jump. Prescription is deliberately strict: only tiers QUOTED in the message and inside `[low, current)` count — bare prose ("too high", "none of the…") keeps the one-rung walk, and never-retryable tiers are structurally excluded, so no message shape can suppress recovery below the status quo.
- **The claude_code advisory gateway forwarded any requested effort verbatim into the SDK.** `ultra` is a codex-family tier; the claude surface tops out at `max`. The gateway now clamps above-`max` requests at the boundary and disclosure is emitted where it is visible and true: the parent process logs the value adaptation (the child's stdout JSON never carried the child-side log), and the child logs only when effort actually survives into the constructed options, so the disclosure can no longer claim a value an older SDK silently omitted. Tiers at or below `max` pass through untouched.
- **The terminal-bench adapter's `--review-effort` choices were a stale subset** and now derive from `EFFORT_SCALE`. The web-search tool's `low/medium/high` enum was examined and deliberately left alone: its send path is a raw `client.responses.create` with no wire recovery, so widening it would convert accepted input into provider 400s rather than adapted requests.

## Known limitation, left as a maintainer decision

A provider that rejects an unsupported effort while quoting ONLY the rejected value (no supported-values list) leaves the jump with nothing to target, so a cold route under the two-send acceptance/P3 rail still cannot walk `ultra → max → xhigh` — the same pre-existing recovery-depth class that already affected `max` against a high-topped provider, now with one more instance. Widening the rail to three sends was prototyped and reviewed, and deliberately dropped: the physical-attempt counter is shape-blind, and on providers that wrap errors in HTTP-200 bodies WITH reported usage a third send can become a third paid settlement — changing the acceptance rail's cost contract is a policy decision this PR declines to make. The alternatives the review panel proposed (withholding `ultra` from review surfaces, or a typed recovery-credit mechanism in usage accounting) were declined as breaking the uniform-scale design and over-engineering respectively. The residual is narrow (owner picks `ultra` on an API review row against a terse-erroring provider) and self-evident in the actor error (`physical attempt limit exhausted`).

## Verification

`node --test web/tests`: 564/564. Full pytest: 11672 passed, 0 failed (one known pre-existing upstream deselect: `test_smoke::test_tool_set_matches`). The per-commit size-ratchet smoke is green at every commit; `request_wire_recovery.py` and `config.py` both stay exactly at their size pins (1000 / 1600 lines). New behavioural pins: the 8-tier scale order with `ultra` on top and `effort_one_step_down(ultra) == max`; `ultra` accepted end-to-end (`resolve_effort`, `normalize_reasoning_effort`, reviewer-slot and subagent-row validation, the legacy migrator); the prescribed-tier jump (quoted ladder → one action, adjectival "too high" and prose "none of…" → one-rung walk, sub-`low` prescriptions excluded); the `switch_model` refusal with no half-applied call; the gateway clamp with `high` untouched.

Six reviewer-panel rounds (claude fable at max, codex sol at ultra — the new tier reviewed itself); every finding was verified against the code before being applied or declined with recorded rationale, and the one concern left standing is the disclosed limitation above.


ouroboros-agent:
Hey @ndrew1337, thank you for this one. The ultra tier and the effort sweep were the right call, and the writeup made review straightforward. While your branch was in flight we retired the Claude runtime on the mainline, so the gateway clamp and its tests lost the file they guarded. I pushed a merge of the current mainline into your branch and let those two pieces go with the retired runtime. Everything else survived as you wrote it. On top I renamed one test whose name went stale, reworded a comment that still told the old clamping story, and added a small guard test so the web effort lists cannot drift from the Python scale again. Before trusting the premise I checked a live Codex CLI 0.150.1 manifest, it does advertise ultra above max. Thanks again for the careful work, delegation to codex lanes is genuinely better for it.

PR #405, opened 2026-08-30
Fix acceptance fence leak: idempotent begin, dead-owner GC, bounded wait

razzant:
During the CyberGym full1507 run (2026-08-30, 32 workers, isolate state on a
network FS) we hit a fence leak: the worker's fence-begin ack is a polled file
with a 10s timeout, and when the ack was missed the supervisor had already
created the fence. The worker had no token, every later begin was rejected
"already active", and the acceptance wait loop burned one paid LLM round per
iteration until the 4h task deadline. 12 tasks spun 365-488 rounds without
ever starting a review panel.

This branch fixes the class:
- begin is idempotent: "already active" for the same root re-adopts the
  existing fence (token + generation); a provably dead owner is reconciled
  with a fresh token
- the ack timeout comes from config SSOT
  (OUROBOROS_ACCEPTANCE_FENCE_ACK_TIMEOUT_SEC, default 120s) instead of a
  hardcoded 10s
- the wait loop is bounded (OUROBOROS_ACCEPTANCE_FENCE_WAIT_MAX_ROUNDS,
  default 3) and then terminalizes as infra_failed instead of spinning to
  the deadline
- dead-owner fences are collected: release in the task reaper after
  confirmed death, a sweep in enforce_task_timeouts, and orphan-heal
  terminalization releases too; reaping/wedged and live direct-chat owners
  count as alive (the predicate consults the reaping registry and
  chat_turn_liveness)

Sealed fences stay non-reopenable; the fence still blocks subtask admission
during review.

Evidence: reviewed by three independent model lanes (GLM, Grok, GPT-5.6-sol)
across the base diff and two fix deltas; all findings dispositioned.
32 fence tests + adjacent suites green, size ratchet green in both modes,
ruff F clean.

Residual: the reaper-exception test uses a synthetic poisoned job; the
late-death path is covered via the real orphaned-running sweep.


razzant:

PR #420, opened 2026-08-31
CyberGym run hardening: DSML fail-closed, container/budget leak fixes, local state dir, reconcile mode

razzant:
Hardening for the CyberGym benchmark harness after the failed full1507 run
(2026-08-30). Ported onto the split cybergym module layout.

Run-path fixes:
- Remote DSML parse is fail-closed: leftover tool-markup finals classify as
  protocol_fail (infra), not capability-missing-poc
- Container leak closed: a timeout after docker run now recovers and removes
  the workspace container (previously 33 -> 192 containers, isolate overload)
- Dead reservations no longer dump the catalog: unresolved upper bounds of
  corpse attempts are excluded from budget liability
- Leftover empty campaign networks are reaped before the next paid start
  (fail-closed when containers are still attached)
- arvo:64622 skipped as infra (broken official pin, dangling symlink)
- Dead-gateway circuit breaker: 3 consecutive transport failures to a dead
  gateway stop dispatch with a typed gateway_unreachable outcome (exit 2)
  instead of burning the remaining catalog into infra_failed rows

Run config: review mode auto (advisory), 400 max rounds, 2h task ceiling,
and the task prompt now shows the agent its wall-time budget derived from
task_timeout_sec.

Isolate state dir: --state-dir keeps the isolated server's state (locks,
logs, task_results) on a local disk instead of the network FS that killed
17 tasks via flock timeouts. Network filesystems are rejected with a clear
error (escape hatch --allow-network-state-dir).

Reconcile mode: --reconcile <run_dir> adopts an interrupted run's gateway
checkpoints, reads terminal results from the isolate disk when the server
is dead, and runs the shared delivery path (submit -> verify -> classify ->
result_index) without re-running agents. Append-only, idempotent under the
result-index lock, concurrent reconcile processes refused, container cleanup
only after the row and settlement are durable, empty/incomplete runs exit
nonzero with a typed outcome.

Evidence: reviewed by three independent model lanes (GLM, Grok, GPT-5.6-sol)
across the port and two fix deltas; all round-1 majors closed and re-verified.
197 cybergym tests + adjacent suites green, size ratchet green, ruff F clean.

Disclosed residuals (accepted, to be fixed separately):
- result_index append has no fsync (pre-existing shared seam across all
  benchmark adapters)
- running reconcile concurrently with a still-live original launcher can
  double-submit identical PoC bytes (operator error; the verifier dedups;
  reconcile is documented for interrupted runs)
- a LedgerError from settlement caused by ledger corruption (vs an inactive
  claim) is misclassified and can delete the container
- the state-dir FS check is fail-open for exotic/unknown filesystem types
  (deny-list covers ceph/nfs/cifs/fuse.ceph)
- a mixed pending+unaccounted reconcile exits 2 but labels the manifest
  outcome "reconciled" instead of "reconcile_incomplete"
- completed results from attempts whose workspace was reaped after a
  post-admission timeout are undeliverable in reconcile (accepted tradeoff)


razzant:

PR #424, opened 2026-08-31
Web UI: dialogs above every overlay; bounded per-task chat ledgers

### LINKED ISSUE #135: Main-chat UI: liveCardRecords and pendingUserBubbles grow unboundedly on long sessions
Tracked by Ouroboros (perf/lifecycle sprint).

The sprint gave project chat panels a full lifecycle (destroy on switch, growth caps: persistedHistory 200, pendingSuggestedNames 100, cancelableTaskIds pruned on every terminal path). The always-alive Main instance still accumulates liveCardRecords/pendingUserBubbles over very long sessions. Apply the same caps and terminal pruning to the Main instance.

### LINKED ISSUE #146: Dialogs render under the reconnect overlay (z-index 90 vs 1000)
Pre-existing: in-app dialogs use z-index 90 while the reconnect overlay uses 1000 (web/style.css:2745 and :3190 at 111bd456), so an open dialog — including the Panic confirm — is covered if the WS reconnect overlay appears. Decide layering policy (dialogs above overlay, or overlay suppresses dialogs explicitly). Sprint tails ledger #6.

kazzand:
Fixes #146, fixes #135 (quick-wins sprint, web-UI stream).

- **#146**: `.confirm-dialog-backdrop { z-index: 1100 }` — confirm/input/alert dialogs (Panic confirmation included) now stack above the reconnect shade and the update overlay instead of rendering under them, visible but unclickable. Content modals deliberately stay under the shade. Static z-order test pins the layer scale; disclosed side effect: toasts (120) render under an open dialog.
- **#135**: new `web/modules/chat_ledgers.js` bounds the always-alive Main instance's per-task state. `liveCardRecords` capped at 200 — the oldest *finished* card is evicted together with its DOM node (in-repo ephemeral-conversion precedent; reload-parity for the user), cascading its task-keyed satellites, retiring the id, and dropping the review hydrator's per-task state. The job-keyed `skillReviewDetailStore` (heavy payload cache) is bounded at its write seam via a facade, so click-driven loads are capped too. `recordConcludedActivity`/`recordTerminalActivity`/`scheduleTaskUiCleanup` moved verbatim; `chat.js` net-shrinks under its shrink-only byte ratchet (224244 → 224061 bytes; manifest regenerated). `pendingUserBubbles` deliberately NOT capped — verified non-leak, and a cap would break the delivered-transition (details in the commit message).

## Testing
705 web node tests (incl. new unit + DOM-integration suites driving 210 sealed tasks and a settle-over-cap scenario), `tests/test_web_dialogs_static.py`, size-ratchet lane — all green.

## Status

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -19,7 +19,7 @@ launcher.py (PyWebView)       ← desktop window, release-reviewed outer shell (
 server.py (Starlette+uvicorn) ← HTTP + WebSocket on configurable host:port (default localhost:8765; Docker/non-loopback supported via OUROBOROS_SERVER_HOST=0.0.0.0)
   │
   ├── web/                     ← Web UI (SPA with ES modules in web/modules/)
-  │   └── modules/review_presentation.js, review_dom_patch.js, and harness_presentation.js ← Review Checkpoint grouping/status, keyed DOM reconciliation, and neutral harness identity presentation; read-side only
+  │   └── modules/review_presentation.js, review_dom_patch.js, and harness_presentation.js ← Review Checkpoint grouping/status, keyed DOM reconciliation, and neutral harness identity presentation; read-side only. modules/chat_ledgers.js ← bounded per-task presentation ledgers for a chat instance (live-card cap with finished-only DOM-and-record eviction, job-keyed review-detail FIFO, terminal-activity keepers) — extracted from chat.js at its shrink-only byte-ratchet boundary
   │
   ├── supervisor/              ← Background thread inside server.py
   │   ├── active_activity.py   ← In-memory thread-safe registry (`DirectActivityRegistry`, `track_direct_activity`) for in-flight direct and ephemeral chat turns; feeds `/api/state` (`active_direct_turns`) and WS typing frames with `activity_id`, `client_message_id`, `phase`, and `kind` without creating supervisor queue records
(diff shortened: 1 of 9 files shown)


ouroboros-agent:
Thank you, and sorry for the amount of rework on this one. The dialog fix is exactly right and it landed as you wrote it, including the layer scale comment and the static test; we also verified it live in headless Chromium, the Panic button now receives a real click while the reconnect overlay is up.

The memory bound took a different shape. The eviction module had two defects we could not ship: a card that finished and was evicted in the same frame armed a 120 second timer nothing could cancel, which made every web test run take two minutes, and during a history rebuild the batch mounted evicted nodes back into the feed without their records. Rather than patch both, the landing reuses the transaction the chat already performs on every reconnect: past 200 live cards the next history sync fully rebuilds from durable history. That is about a dozen lines instead of five hundred, the web suite is back to six seconds, and the lineage maps and cleanup timers are cleared by the same block that always cleared them. Two things to know: a Load older window is the owner's, and a rebuild triggered under it replays that wider quota rather than cutting it back; and the bound is relative, counted against the population the last rebuild left behind, because a history window mints cards itself (summary rows, progress rows, lineage rows) and can hold more than 200 on its own. An absolute cap would have rebuilt on every later sync. The bounded skill review detail cache stayed, as a one line FIFO trim at the store's single write site.

Reviewed on the exact landing tree with the repository's own commit gate (evidence at review_runs/sprint_pr424b_20260902T220443Z in the maintainer workspace; triad clean, scope 8 of 8 PASS). Its first pass blocked this lane twice and both findings were fixed rather than argued: an arm raised while a routine replay was running got discarded at the end of that same sync, so growth arriving only through history stayed unbounded, and the visible dialog change had no vision inspected evidence. There is now a regression that fails without the fix, and a headless Chromium run over this tree with the reconnect shade up, inspected by eye. Quick test and betterleaks green. Merged.

One more thing landed right after, in a maintainer follow up rather than in your PR: a final review of the whole sprint found that the arm could be starved by the resync debounce (every completion restarted the 700 ms timer, so a burst of completions kept the rebuild from ever running) and could be answered by a window fetched before it was raised. The arm, the floor and the cap now live together in createLiveCardBound next to the resync scheduler, with unit tests and a deferred fetch regression. Your design is intact; it just has one home now.

PR #425, opened 2026-08-31
Chat plumbing: honest history --limit end to end; delete the dead log poll queue

### LINKED ISSUE #172: CLI chat history --limit is a placebo: the server ignores the limit parameter
## Current behavior

`ouroboros chat history --limit N` (default 100) sends `limit=N` to `GET /api/chat/history`, but the endpoint accepts the parameter and ignores it — `_int_param` is never called for `limit`; a code comment marks it "accepted for backward-compat but no longer governs the slice". The CLI therefore always receives the server-default window regardless of the flag. The web client's `?limit=1000` is the same placebo.

## Impact

Cosmetic-to-misleading: `--limit 10` still downloads and prints the full default window; `--limit 5000` does not raise it. No correctness issue, but the flag lies about what it does.

## Proposed fix

Map `--limit` onto the real server quota (`n_human`), clamped to the server cap, and drop the dead `limit` parameter from the web client URL. Deferred out of the current performance PR by owner decision (2026-08-09) to keep that PR focused; the change is small but the CLI surface deserves its own review.

Filed as follow-up from the 2026-08-08 performance sprint research (R3).

### LINKED ISSUE #356: Remove dead LocalChatBridge._log_queue/ui_poll_logs drain
supervisor/message_bus.py::LocalChatBridge.push_log still enqueues every event into _log_queue (maxsize=1000, drop-oldest), but its only reader ui_poll_logs has zero callers repository-wide (verified during PR #353 review by two independent reviewers). The queue is dead weight on every live log event; A2A events also enter it before the broadcast guard. Delete _log_queue and ui_poll_logs (or document why they must stay).

kazzand:
Fixes #172, fixes #356 (quick-wins sprint, chat-plumbing stream).

- **#172**: `GET /api/chat/history` ignored the legacy `limit` parameter while every shipped CLI kept sending it — a placebo flag. Closed on both sides: the server now honors legacy `limit` as the *default* for `n_human` (explicit `n_human` wins), the CLI requests `n_human` directly (default raised 100 → 150 to match the server's own window), and the caller-less `api_client.chatHistory()` helper is deleted outright.
- **#356**: `LocalChatBridge._log_queue` (1000-slot drop-oldest churn on every event, A2A copies included) fed a poll drain — `ui_poll_logs` — that has zero callers repo-wide. All three pieces deleted; the WebSocket broadcast in `push_log` remains the one delivery path. `message_bus.py` drops below the 1001-line band; size-ratchet manifest regenerated.

## Testing
`test_message_bus`, `test_log_forwarding`, `test_cli_entrypoint`, `test_chat_history_quotas` (new legacy-fallback + precedence tests), `test_gateway_history`, size-ratchet lane — all green.

## Status

### DIFF
--- a/ouroboros/cli.py
+++ b/ouroboros/cli.py
@@ -290,7 +290,10 @@ def _chat_send_command(args: argparse.Namespace) -> int:
 
 
 def _chat_history_command(args: argparse.Namespace) -> int:
-    _print_json(_client(args).request("GET", f"/api/chat/history?limit={int(args.limit)}"))
+    # `n_human` is the explicit quota for the conversation (non-progress) rows;
+    # the server separately honors legacy `limit` as its n_human default so
+    # already-shipped CLIs are covered too.
+    _print_json(_client(args).request("GET", f"/api/chat/history?n_human={int(args.limit)}"))
     return 0
 
 
@@ -552,7 +555,12 @@ def build_parser() -> argparse.ArgumentParser:
     chat_send.add_argument("text", nargs=argparse.REMAINDER)
     chat_send.set_defaults(func=_chat_send_command)
     chat_history = chat_sub.add_parser("history")
-    chat_history.add_argument("--limit", type=int, default=100)
+    chat_history.add_argument(
+        "--limit", type=int, default=150,
+        help="conversation rows to fetch (server quota n_human: non-progress rows, "
+             "task summaries included; default matches the server's own 150, "
+             "clamped to the server cap)",
+    )
     chat_history.set_defaults(func=_chat_history_command)
 
     logs = subparsers.add_parser("logs", help="read runtime logs")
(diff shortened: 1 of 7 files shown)


ouroboros-agent:
Thank you, this one was clean. The placebo --limit was a real annoyance and the dead log queue was doing work on every event for no reader.

Two maintainer touches landed on top of your commits, with your authorship kept. The CLI no longer hard codes 150: an omitted --limit sends no quota and the server's own window applies, so the number has one owner. And a non positive legacy limit now falls back to the default window instead of returning an empty conversation, which is what a garbled value already did. The help text no longer names the wire parameter.

One correction for the record, since the PR body says message_bus.py dropped below the 1001 line band: it went from 1241 to 1219 and is still inside the band with its existing rationale. The manifest line the PR removed was about gateway/history.py and was already gone on the tip. Not a problem, just so nobody copies that sentence into a changelog.

Reviewed on the exact landing tree with the repository's own commit gate (triad plus scope review, evidence at review_runs/sprint_pr425_20260902T184336Z in the maintainer workspace; triad clean, scope 8 of 8 PASS), quick test and betterleaks green on the branch. Merged.

PR #426, opened 2026-08-31
Hardening: pin Telegram trust_env=False; bound run_script inputs and safety-check subjects

### LINKED ISSUE #377: TelegramClient should pin trust_env=False
The Telegram client currently inherits proxy settings from the environment. For parity with the LLM no-proxy clients and to avoid surprising proxy-injection on shared hosts, construct its HTTP session with trust_env=False (or an explicit owner-configured proxy). See skills/telegram/lib/telegram_api.py.

### LINKED ISSUE #391: run_script.script/args carry no maxLength
The tool schema for run_script accepts unbounded script and args strings. The safety
supervisor prompt embeds the call subject verbatim, so a single oversized script inflates
the safety check prompt beyond provider limits before any dialog-section capping can
help. A maxLength on the schema (or subject-side truncation with an honest marker) is the
class fix. Found while reworking PR #349.

kazzand:
Fixes #377, fixes #391 (quick-wins sprint, hardening stream).

- **#377**: both `TelegramClient` HTTP paths (`call()` and `_download_bytes()` — bot token in the URL) now pin `trust_env=False`, matching every sibling client in the skill and the runtime's no-proxy LLM clients; the transport test asserts both pins (mutation-tested). Repo-wide audit of remaining env-trusting clients will be filed as separate issues per the agreed report-only scope.
- **#391**: `run_script` accepted unbounded script/args; the safety supervisor embeds the call subject verbatim, so one oversized script inflated the check prompt past provider limits. Two complementary layers, neither truncating what the reviewer sees: (1) schema + handler enforce script ≤ 200k chars, args ≤ 100 × ≤ 10k (typed refusals; the only bound in light/off safety mode); (2) the safety seam refuses any serialized subject over 250k chars that would enter the LLM check — `run_command ["sh","-c",huge]` included — with the typed `SAFETY_SUBJECT_TOO_LARGE_BLOCKED` policy denial plus a durable audit event. Subject rendered `ensure_ascii=False` (non-ASCII no longer 6×-inflated; near-cap Cyrillic test) with lone surrogates substituted. The deterministic safe-subject whitelist and the documented no-backend fail-open trio never build a prompt and are deliberately unaffected (pinned by tests). `ARCHITECTURE.md`'s safety row updated in the same change.

## Testing
`test_telegram_transport_safety`, `test_safety_rate_limit` (subject-budget, Cyrillic, surrogate, whitelist pins), `test_shell_run_shell`, `test_safety_policy`, `test_provider_contract_catalog`, size-ratchet lane — all green.

## Status


ouroboros-agent:
Thank you. The subject budget in the safety supervisor is the right fix at the right seam: one place builds the check prompt, an oversized subject is refused with a typed outcome and an audit event instead of being truncated, and Cyrillic no longer inflates six times on the way to the model. That part lands as you wrote it.

Two things changed on top, with your commits intact. First, the run_script handler caps and the shell_bounds.py module were dropped: the caps ran after the safety call, so a 200 to 238k script paid the check and was then refused for free, and in light or off safety mode no prompt is built at all, so the caps bounded nothing there; the budget at the seam already closes the class and exec size limits have their own home in argv_budget.py. Second, the Telegram pin is now conditional: an install that routes through a proxy keeps it, everyone else is isolated from ambient proxy and SSL environment, mirroring what the LLM lane does. The decision is made once in plugin.py and passed into the client, so the library file stays free of core imports. One residual is disclosed in the commit: an install that trusts a custom CA only through SSL_CERT_FILE and uses no proxy loses Telegram egress until the CA is installed system wide, which is how the menu client in the same skill already behaves.

The two prompt hunks were not taken: the SYSTEM.md paragraph no longer exists on the tip and the checklist forbids restating a structurally enforced gate in a prompt. The ARCHITECTURE row was corrected: fail open cases a and b return before any prompt, case c on the local fallback is now refused before the attempt.

Reviewed with the repository's own commit gate on this lane's landing tree (evidence at review_runs/sprint_pr426_20260902T192941Z in the maintainer workspace; triad clean, scope 8 of 8 PASS). After that gate the branch was updated to the tip, which added the neighbouring lane's content that its own gate had already covered, in files this lane does not touch. Quick test and betterleaks green on the merged head. Merged.

PR #427, opened 2026-08-31
Hygiene: env-leak class closure in tests; truthful README tags claim; real teardown reasons

### LINKED ISSUE #121: README Version History references 6.87.6 and 6.87.2 with no matching git tags
## Problem
README Version History lists 6.87.6 and 6.87.2 and states that history is "preserved in Git tags", but no git tags exist for those two versions. The documented recovery path does not work for them.

## Proposed fix
Either restore the missing tags, or reword the Version History note so it does not promise tags that are absent.

Deferred under the owner's proportionality directive.

### LINKED ISSUE #153: Parameterize the forensic reason in _kill_orphaned_children
launcher.py:657 (at 111bd456): _kill_orphaned_children always records reason 'window_close', including headless keep-alive, startup-abort, startup-failure and panic teardowns — misleading in post-mortem logs. Add a reason parameter and pass the real trigger. Sprint tails ledger #15.

### LINKED ISSUE #282: Test files call apply_settings_to_env without restoring the environment (flake class)
Root cause of the historical test_review_verification_v6544 flake, found during the review-cycles lane of PR #276: several test modules (about eight, grep apply_settings_to_env in tests/) apply settings into os.environ and restore with delenv(raising=False) on keys that did not exist before, or do not restore at all. Under xdist the leaked keys change behaviour of unrelated tests in the same worker (enforcement mode, review slots, model names).

Fix as a class: one fixture that snapshots os.environ before apply_settings_to_env and restores it exactly in teardown, used by every call site; forbid direct apply_settings_to_env calls in tests via a small guard. One instance was fixed in tests/test_settings_file_roundtrip.py (setenv then delenv); the rest still leak.
Origin: PR #276 (managed update flow redesign) review cycles and live end-to-end run, anchored at merge f6bb6b1259cd.

kazzand:
Fixes #282, fixes #121, fixes #153 (quick-wins sprint, hygiene stream).

- **#282**: one autouse conftest fixture snapshots `os.environ` before every test and restores it *in place* (clear + update on the real `os._Environ`) — closing the whole env-leak class instead of policing individual `apply_settings_to_env` call sites. The 10 `monkeypatch.setattr(os, "environ", dict(...))` instances are removed as the same class (a dict swap severs the putenv sync subprocesses inherit). Self-contained canary drives the fixture's own generator.
- **#121**: the README Version History footer promised "Git tags and GitHub releases … full bodies remain at their git tags", but 20 of the 77 enumerated versions have no tag at all. Both clauses rewritten to what is actually true for every listed version: bodies remain in this file's own git history, tags where present. No tags created (maintainer authority); the forward P9 invariant stays strict.
- **#153**: `_kill_orphaned_children` hardcoded `window_close` into the cleanup journal, so panic stops, headless shutdowns, crash-fuse exits and startup failures all read as a closed window. A `reason` parameter now names each trigger (`panic_stop` / `headless_shutdown` / `crash_fuse` / `startup_abort` / `startup_failure` / `window_close`), the headless teardown discriminates shutdown vs crash fuse with the same predicate as its exit code, and the reason is journalled at teardown *entry* — the adversarial review proved `stop_agent()` usually consumes the process record first, so a row-only reason would rarely land.

## Testing
`test_environ_isolation` (new), `test_launcher_headless_fallback` (strengthened per-scenario reason pins + no-record journal test), the six env-touching suites incl. subprocess-spawning `test_packaged_runtime_and_lifecycle`, size-ratchet lane — all green.

## Status


ouroboros-agent:
Thank you. The environment leak class was real: on the base ten tests from twelve suites leaked about a hundred keys between tests, and with your fixture that is zero.

Maintainer touches, your commits intact: the README footer keeps your wording with the tip's 6.110.1 row added; a small test pins that the snapshot fixture really is autouse, since removing the decorator left every test green; the runtime mode fixture lost its now redundant restore; the DEVELOPMENT sentence about monkeypatch was reworded to the reason that is actually true (both mechanisms are per test, monkeypatch reverses exactly the named change); and the headless teardown now reads the shutdown event once before stop_agent, so a signal arriving during the stop cannot label a crash fuse exit as a clean shutdown, with a behavioural test for that window.

Two numbers for the record: the PR removed 14 monkeypatch environ idioms, not 10, and none remain; and the footer sentence was indeed untrue, 22 of the 78 listed versions have no tag, plus two visible rows. That gap is now tracked separately.

Reviewed with the repository's own commit gate on this lane's landing tree (evidence at review_runs/sprint_pr427_20260902T190643Z in the maintainer workspace; triad clean apart from one advisory self consistency note, scope 8 of 8 PASS). Two things came after that gate and are stated plainly: the advisory note was fixed on top (comments and one handbook sentence, no executable change), and the branch was updated to the tip, which added the neighbouring lane's content that its own gate had already covered. Quick test and betterleaks green on the merged head. Merged.

PR #441, opened 2026-08-31
Reconcile abandoned unresolved usage rows so cost_final can close

razzant:
A transient provider error (e.g. a 429) leaves a usage-ledger attempt row in
`unresolved` state forever: `terminalize_abandoned_attempt` existed but had
zero production callers. The stuck row makes `cost_final=False` forever
(`non_final_rows` never clears), and downstream cost-finality gates block on
it. This dropped ~30 solved tasks in a CyberGym benchmark run (2026-08-31):
the adapter refused completed gateway frames whose cost never finalized.

This adds a production reconciliation path so a transient provider failure
can't block cost-finality forever:

- `terminalize_abandoned_attempt` now settles an already-unresolved row at
  its carried reservation upper bound with `cost_final=True`, disclosing the
  write-off via a typed `settle_reason` (money is never understated; a 429
  slightly overstates, which is the intended conservative direction).
  Without a known bound the row stays honestly unresolved.
- The ledger state machine gains exactly one legal exit from `unresolved`
  (to `settled` with the typed reason); the validator pins the write-off
  amount and finality (cost_usd == carried bound, cost_final == True), any
  other unresolved transition stays UsageLedgerCorrupt.
- Reconciliation runs on two existing seams (no new scheduler): task
  terminalization (reconstruct_task_cost reconciles the task's own + subtree
  rows regardless of age, so the terminal frame is born cost-final) and the
  periodic supervisor maintenance tick (rows older than the TTL).
- TTL is `OUROBOROS_USAGE_UNRESOLVED_WRITEOFF_SEC` (default 900s, clamped
  60-86400) from config SSOT.

Evidence: reviewed by three independent model lanes (GLM, Grok, GPT-5.6-sol);
the one major (validator didn't pin the write-off amount) is fixed and
re-verified. 531 tests green across the usage/accounting/custody suites,
size ratchet green, ruff F clean.

Disclosed residuals (separate owner decisions, not this PR):
- orphaned `dispatched` rows (worker killed between dispatch and outcome)
  also block cost_final forever, but a naive TTL sweep can't tell a dead
  dispatch from a slow live provider call (TTL 900s < per-call ceiling
  1800s) — needs a liveness-aware design, separate decision.
- reconciler errors in the two seams are swallowed at DEBUG (fail-open so a
  reconciler fault never breaks cost reads); whether to surface them as
  owner-visible alerts is an observability policy decision.


razzant:

PR #456, opened 2026-09-01
Consolidate CyberGym recovery and accounting hardening

razzant:
Consolidates the CyberGym recovery work from [#405](https://github.com/razzant/ouroboros/pull/405), [#420](https://github.com/razzant/ouroboros/pull/420), and [#441](https://github.com/razzant/ouroboros/pull/441) on top of current ouroboros.

This makes acceptance fences bounded and recoverable, reconciles abandoned usage accounting, and makes CyberGym delivery crash-resumable across result recording, claim settlement, workspace cleanup, retries, and launcher/reconcile ownership. It also enforces local mutable state, deterministic gateway circuit breaking, and fail-closed DSML handling.

Owner update: the full 1,507-task campaign is authorized at 64 fixed lanes. The first 64 terminal rows are a rolling observation checkpoint, not a task-65 barrier; admission continues until monitoring latches a stop.

Validation through da8afcdb: the CI-shaped non-serial suite reached 100% with exit 0 on 010c4d2e; subsequent 64-lane, sparse-recovery, and spawn-safe server deltas passed focused tests, tree-wide Ruff, authoritative size-ratchet, and focused review. The serial suite previously reached 100% on the byte-identical core at d460635d.

The benchmark launch is managed separately from this PR, and this PR is not authorized for merge.

Update 2026-09-03: the remaining-1150 cohort (6bab3eab) died to the same supervisor-loop stall class as full1507 — py-spy caught the loop parked in `_handle_task_heartbeat` → `live_root_cost_projection` → `_memoized_final_rows` → `_locked` on `usage_attempts.lock` (three dumps; stalls of 90–105 s past the liveness deadline), the llm_usage budget refresh and the `assign_tasks` pre-check share the shape, and `api_task_get` ran the same locked read inline on the asyncio loop, so the executor's 60 s status polls and 30 s cancels died at transport level (67 GatewayTransportError write-offs, then the cancel/crash storm).

- 7c0df784 fixes the class at the substrate: the ledger-lock acquire raises a dedicated `UsageLockUnavailable`, and the rows memo grows a stale-while-revalidate path — display readers (heartbeat cost projection, llm_usage budget refresh, assign_tasks/chat/evolution budget pre-checks, `/api/state`, the `/api/tasks/{id}` cost view, the cost-breakdown endpoint) pass a sub-second timeout and serve the last validated snapshot under contention, with a 1 s revalidation backoff; a cold memo still fails closed, and `reserve_attempt` stays the exact fail-closed monetary gate. `api_task_get` and `api_cost_breakdown` also move their reads off the event loop.
- Proof: `tests/test_usage_projection_contention.py` runs the real monetary write path from 64 threads (the production lane count) against a real on-disk ledger, proves the convoy is real (58/60 non-stale short-timeout reads lose the lock), and drives the real loop/gateway readers through it without a single read crossing 5 s. The 64-lane acceptance-fence battle test stays green.
- Earlier follow-ups already on this branch: acceptance-fence IPC recovery (2badc9d0), deferred task_done finalization (060be342, hardened in 6bab3eab), gateway transport ride-out on poll/cancel (d11ec295), query-poc 404 reuse handling (0b3176c9), per-event budget-projection coalescing (b1b6e481), deadline cancel custody (24e47fd7).

Validation on 7c0df784: focused suites (usage accounting/memo/render-cache/reconcile, acceptance fence + 64-lane contention, gateway parity, budget limits/tracking, events/llm_usage, the new contention smoke) all pass; tree-wide Ruff on the changed files is clean. The full non-serial suite shows only the three pre-existing `test_v674_acceptance_dialogue.py` failures that reproduce identically on the clean branch tip in this environment (acceptance-dialogue re-drive, no ledger involvement).


razzant:
Recovery pin `9b7eb684` adds:
- bounded custody-GET retries for transient transport/503 failures;
- tracked draining of deferred task finalizations plus synchronous pending child-ref promotion before attestation/cleanup;
- append-only, hash-linked late delivery for settled executor transport failures, with 6-decimal accounting parity and torn-pair repair.

Verification: 165 affected tests passed, including server-shutdown ordering, custody, reconcile, observability GC, task-done ordering, and usage-ledger contention.

razzant:
**Pin 34feecaf — root-cause fix for the r7 (full1507) collapse, per the three scout reports.**

Root cause, confirmed by 37 py-spy dumps + events timeline: control-plane blob promotion (`gunzip → sha256 → gzip → gunzip → sha256`, ~55k files / 18 GB per cohort) ran on the 4-thread `task_done` finalization pool, the cancel-custody pool, and `supervisor-main`. Lanes never freed → dispatch starvation → submit-anchored 2 h deadline cancelled queued-then-started tasks → 3 transport rows from a ~100 s stall tripped the breaker → 1360 tasks abandoned. Ledger-lock contention (fixed in 7c0df784) was not the r7 cause: 0/7 r7 dumps show `supervisor-main` in the lock.

Changes:
- `ouroboros/observability.py`: `write_blob` stamps `portable` + `compressed_sha256`; `promote_blob_ref` relocates portable blobs by hard link / checked byte copy with zero decompression. Tamper/scope/missing/legacy verdicts unchanged.
- `cybergym_lifecycle.py`: task deadline anchored at first observed non-queued status (queue-wait cap separate; +300 s grace behind the isolate abs ceiling); checkpoint discloses `deadline_basis`/`observed_start_at`.
- `cybergym_dispatch.py`: breaker pauses admission and probes `/api/health` on backoff (30/60/120/300 s), resumes on first answer; only a 1 h exhausted pause opens the circuit. History in `dispatch_events.jsonl` and `extra.gateway_circuit.pause`.
- Tests: portable-promotion contracts; 64-lane copy-back battle test through a production-width pool (zero inflation, p-worst < 5 s); deadline anchoring; breaker pause/resume/open; fixed the pre-existing `test_reconcile_task_defers_workspace_release_until_durable` (adoption now requires a running container).

Not done on purpose: early lane release before durable validation — it would violate the AR2-3 lifecycle-fault law (`RUNNING` must not be released on an unvalidated `task_done`). Cheap finalization makes it unnecessary; the Tier-2 load run on this pin is the gate before any paid relaunch.

Pre-existing on the branch and unrelated: `test_acceptance_fence_contention` fails on this loaded host at baseline too (1 s op timeout under load avg ~20).

razzant:
## Tier-2 load reproduction (real isolate, 64 native lanes, stub provider) — 3 more durable fixes

Harness: the real `server.py` booted from the pinned clone (same supervisor / worker processes / `task_done` finalization pool / observability promotion as the CyberGym isolate), 64 lanes, a local OpenAI-compatible stub provider producing 120 KB incompressible tool output per round, 128 tasks over two waves, 16 never-finishing tasks for the deadline/cancel wave. `py-spy` sampled the server every 6–10 s. Metric added: **finalization lane hold** = worker `task_done` → `artifact_finalized_at` (the time a lane stays occupied after the agent finished).

### What 34feecaf actually did under load
Gateway stayed healthy (GET p99 0.27 s, zero transport errors, zero `supervisor_loop_stall`), but **finalization held every lane 150–400 s (p50 160 s)**. Blob store: 4592 hard-linked (fast path) blobs carrying **0.09 GB**, vs 2352 slow-path blobs carrying **3.33 GB** — every LLM-request payload. Two false positives in the portability heuristic:

1. `8c8e75f7` — the bare `observability/` marker matched the system prompt's drive-layout documentation. Replaced by the ref path shapes `observability/blobs/` and `observability/calls/` (the only paths `write_blob`/`persist_call` mint; `/` is never JSON-escaped, so the marker survives embedding in service result strings).
2. `02b99c71` — the drive-root substring guard matched the prompt's plain-text mention of the agent's own data directory. Removed: `_rewrite_child_ref_tree` rewrites only ref dicts and `task_source` refs, all of which leave one of the four byte markers, so the markers are the whole test.

Result on 64 tasks: lane hold **p50 141 s → 12 s**, p90 18 s; cohort wall time 427 s → 190 s. Tier-1 battle test now seeds production-shaped request payloads (system prompt with layout doc + drive root mention + growing history), so both false positives are covered.

### Found by the deadline-cancel wave
3. `b426202e` — a cancel killed a worker that held `usage_attempts.lock`; the lock stayed orphaned for the full 90 s stale window and **every ledger transaction behind it failed: 17 accounting errors, 10 healthy tasks written off as `infra_failed`** from one cancel. `acquire_exclusive_file_lock` gains `reclaim_dead_owner` (owner pid provably gone via `ProcessLookupError`, inode re-check before unlink; live/unknown owners keep the stale-age contract); the ledger lock opts in. Re-run: 2 of 16 cancels hit a lock holder, both reclaimed immediately, **0 `UsageLockUnavailable`**, 112/112 completed + 16/16 cancelled.
4. `c236e4fc` — `_record_worker_pids` re-fingerprinted all 64 slots (2 × `ps` each) on every respawn, holding the synchronous cancel HTTP response ~40 s on a busy host (> the launcher's 30 s cancel timeout) and appending 64 duplicate custody rows. Each (slot, pid) is now ledgered once per generation.

Final validation run (pin b426202e, 128 tasks + cancel wave): all verdicts green — no stalls, GET p99 0.21 s, 0 transport errors, all completed/cancelled, lane hold p50 16 s / p90 27 s. A short cancel-path check on c236e4fc is running; the full 1507 relaunch goes on that pin.

razzant:
## r8 (pin c236e4fc, full 1507) — first hour, and a dispatch finding

r8 is running on the validated pin: 64 lanes, gateway healthy, 0 `supervisor_loop_stall`, 0 transport errors, 0 ledger-lock failures. Two observations from the live launcher:

**1. Admission was head-of-line blocked (pre-existing since a5e93b86).** `run_dispatched` gated admission on `in_flight + completed < max_workers`, where `completed` holds rows already finished but buffered for source-order recording behind an earlier, still-running position. Every window of 64 tasks therefore waited for its slowest member; with a 2 h task deadline one straggler idles up to 63 lanes for up to 2 h. r8 at +60 min: 23 tasks finished, **0 refills**, 23 lanes idle, position 0 still running (py-spy on the launcher: pool threads parked, the rest polling the gateway). r7 and every earlier 64-lane cohort ran this way — it is the main reason the campaigns were so slow, independent of the server-side stalls.

`cbfb4982` (pushed, **not deployed to r8**): admission is bounded by lanes actually running; rows and `result_index.jsonl` still land strictly in source order; the breaker now sees rows as they settle, so a later transport failure pauses admission immediately instead of waiting behind the head of the line. Two new tests pin both properties; METHODOLOGY gets one paragraph. Whether to restart the cohort on this pin is Anton's call.

**2. 3–11 min between `task_done` and `artifact_finalized_at` is configured post-task cognition, not a stall.** Finalize-pool threads were busy in 3 of ~150 py-spy samples (portable relocation), supervisor-main was idle half the time. For tasks with a `workspace_root`, `_dispatch_root_post_task` runs review/reflection blocking (`OUROBOROS_REVIEW_MODELS`, `effort=max`, up to 2 cycles per `settings_base.json`) before the worker flushes `task_done`. It holds the lane and costs a review round per task, but it is the configured agent, so left as is.

razzant:
## r8 (c236e4fc, full-1507, 64 lanes) — first 2h wave findings and three new pins

Run `c236e4fc-deepseek-full1507-64-20260904T035952Z`, launched 04:00 UTC. At 06:43 UTC: 66/1507 recorded — 46 completed (36 official pass), 20 infra_failed, $72.68. Server never stalled, no transport errors, cancel POST ~1 s (pid-ledger-once held).

### What broke at the 2h wall (06:04–06:15 UTC)
Agent loops finished **inside** the deadline; then the blocking post-task cognition chain (consolidation → summary → reflection) ran 5–16 min under 64-lane load, crossed `deadline_at`, and the reaper killed the worker mid-reflection (120 s finalization grace).
- **8 solved tasks → `ExecutorFailure: cancellation custody did not settle`.** The in-flight `post_task_synthesis` ledger row stayed `dispatched`; the completed frame was born `cost_final:false`; the existing write-off covered only `unresolved` rows, and the launcher's grace path was ineligible (`unresolved_upper_bound_usd = 0`, `cost_estimated` absent). Server-side these tasks are `completed` with PoC + cost.
- **4 tasks → `task_done_lifecycle_fault`.** The worker's own `task_done` raced the reaper's RUNNING pop; root row still `scheduled`, refused. 3 of 4 have a completed child-drive result.
- **2 tasks → `context_overflow`.** DeepSeek V4 Flash emitted single turns with **1113** and **872** `run_command` calls (90K completion tokens); executing them pushed the transcript to ~2M tokens. 8/68 tasks had turns with >20 calls.

### Pins pushed to this branch
- `cbfb4982` — admission counts running lanes only (head-of-line: 36 tasks done, 3 lanes refilled).
- `69e3048a` — `OUROBOROS_MAX_TOOL_CALLS_PER_TURN=32`; surplus pruned from the assistant turn (no orphan tool_call_ids), `tool_call_burst_truncated` checkpoint, model notified.
- `9f2aa2a2` — `OUROBOROS_POST_TASK_COGNITION_MIN_REMAINING_SEC=900`: paid post-task chain skipped when `deadline_at` is nearer than 15 min (`post_task_cognition_skipped` audit row, checkpoint terminal at once). Terminal cost authority now settles the task's **own** open `dispatched` rows at their carried bound and releases `reserved` ones (subtree rows keep the unresolved-only rule), so frames are born final.

Note: `9f2aa2a2` changes agent behaviour for tasks finishing in the last 15 min (no reflection/memory work; the PoC is already final by then). Relaunch on the new pin (r9) awaits Anton's decision; r8 continues on `c236e4fc` meanwhile.

Watch item: 2× `UsageLockUnavailable` at 06:40–06:42 (ledger 31K rows / 46 MB, ~230 transitions/min). Rows went `unresolved` at bound, tasks survived; convoy scales with ledger size.

razzant:
## r8 follow-up (08:20 UTC): two more durable fixes, pin `fb54a3dd`

**r8 status** (pin `c236e4fc`, 4h20m in): 83/1507 recorded — 57 completed (45 official pass), 26 infra_failed; 146 claims / 116 task_done; 0 supervisor stalls, 0 transport failures, $88.87.

### 1. Usage-ledger lock convoy killed tasks — `86e370cf`
A 3-minute burst at 06:45–06:48 produced 13 `UsageLockUnavailable` and 5 tasks died as `task_exception` (rounds 54–136 of real work lost). Ledger appends never paused (~3 rows/s throughout), so this was pure lock saturation, not a stuck holder. Root cause measured on the live r8 ledger (43K rows):

| transition | before | after |
|---|---|---|
| reserve | 1.43 s | 0.043 s |
| dispatch | 0.223 s | 0.017 s |
| settle | 0.644 s | 0.017 s |

`_append_rows_locked` re-validated the **whole** history on every append (`_validate_records([*records, *new])`, 0.6 s), and `_transition` built final-rows dicts over the full history twice. At ~3 transitions/s × 0.2–0.6 s hold the lock ran at ~100% utilisation and unfair waiters starved past the 45 s timeout. Appends now validate only the new rows against the prefix's per-attempt last-state map (taken from the cached validated read when it covers the same extent, else one derived pass — same rules, same `UsageLedgerCorrupt` outcomes); attempt lookups reverse-scan for the last row. Residual O(#attempts) cost is the global-budget `_summary` in `reserve` (~40 ms at 12K attempts); tracked as a watch item for the 500K-row tail of a full run.

### 2. Lifecycle-fault resolver raced the reaper — `fb54a3dd`
4 completed tasks were published as `task_done_lifecycle_fault` at the 2 h wall: the reaper popped RUNNING, the worker's own completed `task_done` landed in that window, and `_resolve_lifecycle_fault` terminalized it as failed before the reaper's post-kill re-check could mirror the completed child result. The resolver now defers while `reaper_owns_task_row(task_id)` — from the RUNNING pop until the reaper's job finishes publication (a separate set from the acceptance-fence `_REAPING_TASK_IDS`, which still clears at confirmed death so fence reclaim is unchanged).

Tests: `tests/test_usage_rows_memo.py` (+4), `tests/test_reaper_owns_lifecycle_fault.py` (new, 4); ledger/budget/reaper/fence/cancel suites green.

`prepare_r9.sh` now targets `fb54a3dd`. r8 stays running on `c236e4fc` until the pin change is approved.

razzant:
## r8 09:30 UTC: second ledger-convoy burst root-caused — pin `9f6700c8`

**r8 status:** 133/1507 recorded — 91 completed (71 official pass), 42 infra_failed, $124.15; 0 stalls, 0 transport failures.

**The 4 h wave killed 15 more tasks** (08:27–08:31, `task_exception: UsageLockUnavailable` / `PhysicalAttemptPreparationFailed: … UsageLock`), almost all at rounds 1–7 — i.e. the *newly admitted* tasks. This is a second mechanism, distinct from the per-transition cost fixed in `86e370cf`: a process's **first** in-lock read cold-parses and validates the whole ledger under the 45 s lock (~1 s at 47K rows). The wave freed ~48 lanes at once → 48 fresh workers each held the lock for a full cold parse, serialized (~50 s) → every reserve behind them timed out. Ledger appends dropped to 21–50/min with 5–17 s gaps while nothing was being written — the lock was held by readers, not writers.

Fix (`9f6700c8`): `_locked()` seeds the per-process read cache from an **unlocked** snapshot of the row-aligned prefix before acquiring the lock; the in-lock read then resumes from it and parses only the appended tail. Torn/corrupt files are left to the locked full read (which owns quarantine) with a 30 s back-off. Measured on the live r8 ledger (52K rows): cold-reserve lock hold **1.4 s → 0.024 s**. 4 new tests (`tests/test_usage_rows_memo.py`); 257 usage/ledger/budget tests green.

Convoy accounting on r8 so far: 23 of 42 infra failures are `UsageLock*` (2 at 06:18, 5 at 06:46–06:48, 15 at 08:27–08:31, 1 at 09:00). Both mechanisms scale linearly with ledger size, so on `c236e4fc` every further deadline wave will be worse. `prepare_r9.sh` now targets `9f6700c8`; r8 continues on `c236e4fc` pending the pin-change approval.

Also noted: `server.log` is being rotated at tiny sizes (2 KB / 198 B / 739 B files since 08:38) — forensics for r8 after 08:38 live in `events.jsonl` and the journal only. Not a run-health issue; will look at what rotates it.

razzant:
Addendum: the `server.log` collapse is root-caused and fixed in `b1d0b334`. Under fork every worker inherited the server's `RotatingFileHandler` (2 MB × 3) and rotated the *same* file on its own byte count; 64 racing renames produced the 2 KB / 198 B / 739 B files. `worker_main` now swaps the inherited handler for a `WatchedFileHandler` on the same path (only the supervisor rotates) and the supervisor's window is 64 MB × 5. `prepare_r9.sh` → `b1d0b334`.

razzant:
### Pin `d077c1bb` — wire disclosure survives a failed post-response settle; r8 → r9 chained relaunch

**New failure class in r8 traced to the ledger convoy.** 6 tasks (arvo:44199, 61699, 40317, 15431, 64286, 60672) were refused by the executor as `gateway telemetry omitted effort for a served call` after 40–100 productive rounds. Root cause: a response whose `settle_attempt` failed with `UsageLockUnavailable` (08:27–08:31 convoy) was captured `unresolved`; `WireUsageDisclosure.from_candidate` → `validate_physical_wire_attempt` requires `state == "settled"`, raised, and `finalize_wire_response` silently skipped the disclosure. That call then carried no `usage.request_wire.applied_effort`, `served_effort_count < trace_call_count`, and the whole task was written off. Evidence: `cybergym-769df41b…` call 1 (AkashML via OpenRouter, ledger rows `reserved → dispatched → unresolved post_response_accounting_failed:UsageLockUnavailable → settled writeoff`) has `request_wire: null`; its 42 sibling calls all carry it.

Fix (`d077c1bb`): the disclosure accepts an `unresolved` **post-response** capture (the candidate was sent and answered; money stays bound at the reservation). Sends that produced no response still cannot be disclosed; compatibility-profile learning (`WireCompatibilityReceipt`) keeps requiring a settled attempt. Test: `test_disclosure_survives_a_failed_post_response_settle_but_learning_does_not`. Together with `86e370cf`/`9f6700c8` (convoy root causes) this class should not recur; if it does, the ledger is the thing to look at.

**r8 stopped at 11:36Z, r9 started 11:39Z on `d077c1bb` as a chained run (per Anton: old runs are kept, new runs continue the same 1507 rather than restarting).**
- r8 final: 205 rows — 138 completed (105 official pass, 76.1 % of completed), 67 infra_failed, $201.46. ~17 in-flight gateway tasks were orphaned by the stop (launcher has no SIGTERM handler) and are simply re-run.
- Catalog for r9: `tasks_remaining_after_c236e4fc.json` = full 1507 minus the 138 tasks r8 recorded `completed` → **1369 tasks**, sha256 `16dbd796e428…`. Rule: infra_failed and unrecorded tasks are re-run on the new pin; completed rows are never re-run. Audit of the derivation: `r9_remaining_meta.json` (excluded ids, re-run ids, rule).
- Concatenated view = latest row per task across r8 → r9, denominator 1507 (`concat_status.py`). Same model / pin lineage / budget / per-task limits; only the code pin and the catalog subset differ, both disclosed in the run manifest.
- r9 run id: `d077c1bb-deepseek-remaining1369-64-20260904T113923Z`, 64 lanes, delivery/provider/watch/stallwatch units active. First check: 64 claims, 64 workspaces, no errors.

razzant:
### Pin `79dc640a` — incremental finals view on the reserve path (pre-emptive, for the r9 tail)

**Why now.** r9 (`d077c1bb`, 1369 remaining tasks) writes ~280 ledger rows per completed task; at 49 completions the ledger is already 32K rows / 48 MB and will end near 400K rows / ~140K attempts. `reserve_attempt` folded `_final_rows` + `_summary` over the **whole** cached ledger under `usage_attempts.lock` on every reserve, then filtered the finals per root task. Measured on the r8 ledger replicated to 512K rows / 171K attempts: **470 ms per reserve**. At 1-2 reserves/s across 64 lanes that is 50-90% lock utilisation from reserves alone — the convoy of 08:27Z (fixed for its first two mechanisms in `86e370cf` / `9f6700c8`) would return in the last third of r9 through a third mechanism.

**Fix.** `_summary` is now a fold over `SummaryAccumulator` (same per-row arithmetic, signed so a superseded final row can be retracted). `_FinalsView` (in `_usage_rows_memo.py`) keeps the final-rows dict, the accumulator and a per-root index, and applies only the rows appended since the previous reserve. It refolds from scratch on any change of the record list's shape (first/last seq, shrink), every 4096 applied rows to bound float drift, and re-folds the summary whenever a subscription row (non-retractable window max) has been seen. Any exception in the view falls back to the from-scratch fold.

Same 512K-row ledger: **p50 0.02 ms / p99 0.10 ms per reserve**, summary equal to the replay. Tests: property test vs. replay after random appends (multi-root, all transitions), refold triggers, subscription fallback, and a reserve-path test pinning that `reserve_attempt` no longer folds the ledger while keeping global/root budget semantics across a cache refold. 222 tests green across the ledger/budget/finalization suites.

**Rollout.** r9 keeps running on `d077c1bb` until just after its first 2h deadline wave (~13:40Z), when in-flight work per lane is at its minimum; then chain-switch to r10 on `79dc640a` with the remaining catalog derived from r8 ∪ r9 completed rows (same append-only concat rule as r8→r9, nothing deleted). Trigger earlier if `UsageLockUnavailable` appears in r9.

razzant:
## r9 (pin d077c1bb) wave losses root-caused — pins `c44d0b0a` + `0470af63`

r9's first 2 h wave lost 18 paid tasks to two infrastructure classes. Both are closed on the executor/server side; neither was the r8 class (post-task cognition past `deadline_at` — the 900 s guard fired 28 times and held).

### 1. `FinalPocRefused` ×9 → `c44d0b0a` (executor classification)
All nine runs finished under their own steam, 40–90 min before the deadline, with a final message and no PoC marker. The runtime marked `execution.status=degraded` because the model had emitted malformed `run_command` argument JSON (`TOOL_ARG_ERROR`) along the way; the executor's fair-completion predicate accepted only `execution.status=ok`, so the model's own mistakes were typed infrastructure and the tasks were re-run — a second attempt the methodology does not grant.

`_gateway_fair_completion` now returns `(verdict, basis)`: a run degraded **solely** by agent-attributable tool errors (`TOOL_ARG_ERROR`, `STR_REPLACE_ERROR`, `LIST_FILES_ERROR`, `TOOL_TIMEOUT`) is a fair `failed`; any other execution status or any tool error outside that allowlist stays unfair and names why. Rows carry `fair_completion_basis`. METHODOLOGY §fair completion updated; 3 new tests. Chain rule updated accordingly: `completed` and `failed` rows are terminal across the chain, only `infra_failed` is re-run.

### 2. `cancellation custody did not settle` ×9 → `0470af63` (launcher + server)
Forensics (all 9): worker terminal 1–90 s **before** `deadline_at`, `task_done` + `cost_final` recorded; then `finalize_task_artifacts → write_workspace_patch_artifacts` took 8–25 min — one `git diff --no-index --numstat` **subprocess per untracked file** (a fuzzing task left 51 678 corpus files on CephFS) on the 4-thread `task_done` pool, so even a 2-file task queued 8 min. Meanwhile the gateway projected `status=running, artifact_status=finalizing`; the launcher cancelled at `deadline_at+300 s`; the cancel path re-ran the same finalization synchronously in the HTTP thread; POST #1 timed out at 30 s, the retry hit `claim_refused → 503`, and the 300 s custody window expired. The server terminalized all nine 1–15 min later (`cancel_intent outcome=already_settled`). py-spy dump 14:07:42Z: `ThreadPoolExecutor-0_7` in `_git_bytes ← _untracked_blob_exclude_reason ← write_workspace_patch_artifacts ← cancel_task_custody`.

Fix, both sides:
- **Launcher** (`cybergym_wire._gateway_finalizing`): a frame with the worker terminal (`child_status`) and `artifact_status=finalizing` is a finished, paid result in flight — the wait loop grants a single 30 min finalization grace (`deadline_basis=finalization_grace`) instead of cancelling, and the post-cancel custody poll keeps custody for the same grace when it sees such a frame (`custody_basis=finalization_grace`). Cancellation of a task that outlives the grace is unchanged.
- **Server** (`headless.py`): the binary verdict for untracked files is computed in-process (NUL within git's first 8000 bytes — git's own `buffer_is_binary` rule) instead of a subprocess per file; the untracked files a patch carries are budgeted (`OUROBOROS_PATCH_MAX_UNTRACKED_FILES=400`, `OUROBOROS_PATCH_UNTRACKED_TIME_BUDGET_SEC=120`), surplus disclosed under `untracked_excluded` with a `untracked_budget` diagnostic and `counts.untracked_budget_excluded`; the patch still writes (no `errors`), so `artifact_status` cannot sit in `finalizing` for tens of minutes. The PoC is read from the workspace, not the patch, so scoring is unaffected.

Tests: 3 new launcher tests (grace on the wait path, granted once, custody outlasts its window while finalizing) + 3 new headless tests (no git subprocess in binary detection, file budget, time budget); affected suites 341 passed.

### Runs
- r10 (pin `79dc640a`, 14:09Z) confirmed the finalization bottleneck live and at a worse scale: 100 `task_done` in the first 40 min (finalization hold p50 473 s / p90 746 s / max 1114 s), yet only 2 rows ever reached `result_index.jsonl` — the 4-thread finalization pool fell hours behind, so the launcher saw `running/finalizing` for nearly every finished task until the 16:10Z deadline wave. Stopped 16:21Z; run root, 171 task dirs and server state kept (reconcilable later; the ledger view `79dc640a` itself held: 0 lock errors).
- r11 launched 16:24Z on `0470af63` with the remaining catalog (1321 = 1507 − every `completed`/`failed` row in r8∪r9∪r10, sha `29fd550c…`). Chain view before r11: 205 recorded, 184 completed, 124 pass (67.4 % of completed), $265. First thing to watch on r11: finalization hold and rows/hour.

razzant:
## r9 (pin d077c1bb) wave losses root-caused — pins `c44d0b0a` + `0470af63`

r9's first 2 h wave lost 18 paid tasks to two infrastructure classes. Both are closed on the executor/server side; neither was the r8 class (post-task cognition past `deadline_at` — the 900 s guard fired 28 times and held).

### 1. `FinalPocRefused` ×9 → `c44d0b0a` (executor classification)
All nine runs finished under their own steam, 40–90 min before the deadline, with a final message and no PoC marker. The runtime marked `execution.status=degraded` because the model had emitted malformed `run_command` argument JSON (`TOOL_ARG_ERROR`) along the way; the executor's fair-completion predicate accepted only `execution.status=ok`, so the model's own mistakes were typed infrastructure and the tasks were re-run — a second attempt the methodology does not grant.

`_gateway_fair_completion` now returns `(verdict, basis)`: a run degraded **solely** by agent-attributable tool errors (`TOOL_ARG_ERROR`, `STR_REPLACE_ERROR`, `LIST_FILES_ERROR`, `TOOL_TIMEOUT`) is a fair `failed`; any other execution status or any tool error outside that allowlist stays unfair and names why. Rows carry `fair_completion_basis`. METHODOLOGY §fair completion updated; 3 new tests. Chain rule updated accordingly: `completed` and `failed` rows are terminal across the chain, only `infra_failed` is re-run.

### 2. `cancellation custody did not settle` ×9 → `0470af63` (launcher + server)
Forensics (all 9): worker terminal 1–90 s **before** `deadline_at`, `task_done` + `cost_final` recorded; then `finalize_task_artifacts → write_workspace_patch_artifacts` took 8–25 min — one `git diff --no-index --numstat` **subprocess per untracked file** (a fuzzing task left 51 678 corpus files on CephFS) on the 4-thread `task_done` pool, so even a 2-file task queued 8 min. Meanwhile the gateway projected `status=running, artifact_status=finalizing`; the launcher cancelled at `deadline_at+300 s`; the cancel path re-ran the same finalization synchronously in the HTTP thread; POST #1 timed out at 30 s, the retry hit `claim_refused → 503`, and the 300 s custody window expired. The server terminalized all nine 1–15 min later (`cancel_intent outcome=already_settled`). py-spy dump 14:07:42Z: `ThreadPoolExecutor-0_7` in `_git_bytes ← _untracked_blob_exclude_reason ← write_workspace_patch_artifacts ← cancel_task_custody`.

Fix, both sides:
- **Launcher** (`cybergym_wire._gateway_finalizing`): a frame with the worker terminal (`child_status`) and `artifact_status=finalizing` is a finished, paid result in flight — the wait loop grants a single 30 min finalization grace (`deadline_basis=finalization_grace`) instead of cancelling, and the post-cancel custody poll keeps custody for the same grace when it sees such a frame (`custody_basis=finalization_grace`). Cancellation of a task that outlives the grace is unchanged.
- **Server** (`headless.py`): the binary verdict for untracked files is computed in-process (NUL within git's first 8000 bytes — git's own `buffer_is_binary` rule) instead of a subprocess per file; the untracked files a patch carries are budgeted (`OUROBOROS_PATCH_MAX_UNTRACKED_FILES=400`, `OUROBOROS_PATCH_UNTRACKED_TIME_BUDGET_SEC=120`), surplus disclosed under `untracked_excluded` with a `untracked_budget` diagnostic and `counts.untracked_budget_excluded`; the patch still writes (no `errors`), so `artifact_status` cannot sit in `finalizing` for tens of minutes. The PoC is read from the workspace, not the patch, so scoring is unaffected.

Tests: 3 new launcher tests (grace on the wait path, granted once, custody outlasts its window while finalizing) + 3 new headless tests (no git subprocess in binary detection, file budget, time budget); affected suites 341 passed.

### Runs
- r10 (pin `79dc640a`, 14:09Z) confirmed the finalization bottleneck live and at a worse scale: 100 `task_done` in the first 40 min (finalization hold p50 473 s / p90 746 s / max 1114 s), yet only 2 rows ever reached `result_index.jsonl` — the 4-thread finalization pool fell hours behind, so the launcher saw `running/finalizing` for nearly every finished task until the 16:10Z deadline wave. Stopped 16:21Z; run root, 171 task dirs and server state kept (reconcilable later; the ledger view `79dc640a` itself held: 0 lock errors).
- r11 launched 16:24Z on `0470af63` with the remaining catalog (1321 = 1507 − every `completed`/`failed` row in r8∪r9∪r10, sha `29fd550c…`). Chain view before r11: 205 recorded, 184 completed, 124 pass (67.4 % of completed), $265. First thing to watch on r11: finalization hold and rows/hour.

razzant:
## r9 (pin d077c1bb) wave losses root-caused — pins `c44d0b0a` + `0470af63`

r9's first 2 h wave lost 18 paid tasks to two infrastructure classes. Both are closed on the executor/server side; neither was the r8 class (post-task cognition past `deadline_at` — the 900 s guard fired 28 times and held).

### 1. `FinalPocRefused` ×9 → `c44d0b0a` (executor classification)
All nine runs finished under their own steam, 40–90 min before the deadline, with a final message and no PoC marker. The runtime marked `execution.status=degraded` because the model had emitted malformed `run_command` argument JSON (`TOOL_ARG_ERROR`) along the way; the executor's fair-completion predicate accepted only `execution.status=ok`, so the model's own mistakes were typed infrastructure and the tasks were re-run — a second attempt the methodology does not grant.

`_gateway_fair_completion` now returns `(verdict, basis)`: a run degraded **solely** by agent-attributable tool errors (`TOOL_ARG_ERROR`, `STR_REPLACE_ERROR`, `LIST_FILES_ERROR`, `TOOL_TIMEOUT`) is a fair `failed`; any other execution status or any tool error outside that allowlist stays unfair and names why. Rows carry `fair_completion_basis`. METHODOLOGY §fair completion updated; 3 new tests. Chain rule updated accordingly: `completed` and `failed` rows are terminal across the chain, only `infra_failed` is re-run.

### 2. `cancellation custody did not settle` ×9 → `0470af63` (launcher + server)
Forensics (all 9): worker terminal 1–90 s **before** `deadline_at`, `task_done` + `cost_final` recorded; then `finalize_task_artifacts → write_workspace_patch_artifacts` took 8–25 min — one `git diff --no-index --numstat` **subprocess per untracked file** (a fuzzing task left 51 678 corpus files on CephFS) on the 4-thread `task_done` pool, so even a 2-file task queued 8 min. Meanwhile the gateway projected `status=running, artifact_status=finalizing`; the launcher cancelled at `deadline_at+300 s`; the cancel path re-ran the same finalization synchronously in the HTTP thread; POST #1 timed out at 30 s, the retry hit `claim_refused → 503`, and the 300 s custody window expired. The server terminalized all nine 1–15 min later (`cancel_intent outcome=already_settled`). py-spy dump 14:07:42Z: `ThreadPoolExecutor-0_7` in `_git_bytes ← _untracked_blob_exclude_reason ← write_workspace_patch_artifacts ← cancel_task_custody`.

Fix, both sides:
- **Launcher** (`cybergym_wire._gateway_finalizing`): a frame with the worker terminal (`child_status`) and `artifact_status=finalizing` is a finished, paid result in flight — the wait loop grants a single 30 min finalization grace (`deadline_basis=finalization_grace`) instead of cancelling, and the post-cancel custody poll keeps custody for the same grace when it sees such a frame (`custody_basis=finalization_grace`). Cancellation of a task that outlives the grace is unchanged.
- **Server** (`headless.py`): the binary verdict for untracked files is computed in-process (NUL within git's first 8000 bytes — git's own `buffer_is_binary` rule) instead of a subprocess per file; the untracked files a patch carries are budgeted (`OUROBOROS_PATCH_MAX_UNTRACKED_FILES=400`, `OUROBOROS_PATCH_UNTRACKED_TIME_BUDGET_SEC=120`), surplus disclosed under `untracked_excluded` with a `untracked_budget` diagnostic and `counts.untracked_budget_excluded`; the patch still writes (no `errors`), so `artifact_status` cannot sit in `finalizing` for tens of minutes. The PoC is read from the workspace, not the patch, so scoring is unaffected.

Tests: 3 new launcher tests (grace on the wait path, granted once, custody outlasts its window while finalizing) + 3 new headless tests (no git subprocess in binary detection, file budget, time budget); affected suites 341 passed.

### Runs
- r10 (pin `79dc640a`, 14:09Z) confirmed the finalization bottleneck live and at a worse scale: 100 `task_done` in the first 40 min (finalization hold p50 473 s / p90 746 s / max 1114 s), yet only 2 rows ever reached `result_index.jsonl` — the 4-thread finalization pool fell hours behind, so the launcher saw `running/finalizing` for nearly every finished task until the 16:10Z deadline wave. Stopped 16:21Z; run root, 171 task dirs and server state kept (reconcilable later; the ledger view `79dc640a` itself held: 0 lock errors).
- r11 launched 16:24Z on `0470af63` with the remaining catalog (1321 = 1507 − every `completed`/`failed` row in r8∪r9∪r10, sha `29fd550c…`). Chain view before r11: 205 recorded, 184 completed, 124 pass (67.4 % of completed), $265. First thing to watch on r11: finalization hold and rows/hour.

razzant:
## r11 died to a host OOM from one runaway lane — pins `f63578ef` + `0158158c`

r11 (pin `0470af63`) ran ~4 h then collapsed: at 19:59:50Z the kernel OOM-killer fired host-wide (`global_oom`, it also reaped another user's process). The victim was one of our workers — `python3`, **anon-rss 373 GiB, total-vm 1 TB**. The isolate limped ~30 min, then at 20:31:37 the launcher exited (status 2) and systemd's cgroup teardown SIGTERM'd the server and all 64 workers at once; the crash-storm detector fired (`crash_storm_detected crash_count=64`) and the launcher wrote **1152 in-flight/queued tasks off as `infra_failed`** in one shot. The 156 completed + 13 failed r11 rows are banked; the infra_failed set re-runs.

Root cause: nothing bounds a single worker's resident memory. A task whose context/tool-output (or a pathological workspace load) grows without limit can balloon one lane to hundreds of GiB and take the whole 64-lane host down with it. One bad lane = total campaign loss.

### Fix `0158158c` — per-worker memory watchdog
`ensure_workers_healthy` (already the per-tick liveness sweep) now samples each **busy** worker's `VmRSS` from `/proc/<pid>/status` and `SIGKILL`s any lane over `OUROBOROS_WORKER_RSS_LIMIT_MB` (default **24 GiB**; a single task's context + tool outputs stays in single-digit GiB, so 24 GiB catches the 373 GiB runaway ~15× early). The kill surfaces next tick as a signal death, which the existing crash path finalizes as a **terminal infra failure with no retry** (a retry would balloon again). One offender fails alone; the host and the other 63 lanes survive. Idle workers are never checked; explicit limit `0` disables it. Tests: `tests/test_worker_crash_retry.py` (kills only the over-limit busy lane, spares under-limit and idle, honors disable) — 13 passed.

### Fix `f63578ef` — finalization throughput (from the r10/r11 finalization-hold finding)
r10 recorded only 2 rows because the 4-thread `task_done` finalization pool fell hours behind a 64-lane wave; py-spy on r11 caught 3 of 4 finalize threads walking a fuzzing corpus one untracked file at a time (`stat` + head read, plus a `realpath` per file when a scratch manifest exists) on CephFS. Changes: probe the untracked set under a count+time budget (`OUROBOROS_PATCH_MAX_UNTRACKED_PROBES=4000`, `OUROBOROS_PATCH_UNTRACKED_TIME_BUDGET_SEC=120`, surplus disclosed under `untracked_excluded`); skip the per-file `realpath` when there is no scratch manifest; size `_TASK_DONE_EXECUTOR` from `OUROBOROS_MAX_WORKERS` (`max(4, min(16, workers/4))`) instead of a fixed 4. Tests green.

### r12 relaunched
`0158158c` (tip; includes `f63578ef` + `0470af63` grace/finalization fixes + the watchdog). Chain excludes every fair-terminal (`completed`/`failed`) row in r8∪r9∪r10∪r11 = 355; **remaining 1152**, sha `f567c527…`. Live 21:56Z, all 5 units active, isolate booted. Also fixed `prepare_chain.sh` to compare full resolved SHAs (git widened `--short` to 9 chars on a prefix collision, silently aborting the first relaunch under `set -e`).

### Accuracy (unchanged by the crash — the OOM only cost re-runs, not capability)
Union r8∪r9∪r11 before the crash: 226/339 completed pass = **67%**, matching the published DeepSeek-V4-Flash CyberGym ceiling (**76.7%**; r8's first-pass 76.1% is dead-on, later links attack the harder residue). No scoring bug: delivery integrity intact (final.poc sha == submit sha, 0 empty), submit-time and official `vul_exit` agree, no broken fix-binary, no poc-hash contamination.

### Open, tracked
- A separate `UsageLedgerCorrupt: attempt … changed after terminal state` at 19:16 (before the OOM) — ledger-integrity class, under investigation.
- r10 (~160) and r11 finished-but-unrecorded gateway results are reconcilable from their state dirs (avoid paying twice).

PR #458, opened 2026-09-01
Add DeepSeek as a first-class direct provider

kazzand:
## Summary

DeepSeek becomes the seventh registered direct provider (`deepseek::` prefix, `DEEPSEEK_API_KEY`, fixed official endpoint `https://api.deepseek.com/v1`), wired through the existing provider registry SSOT (`provider_models.py`) with zero new abstraction layers, following the MiniMax onboarding precedent.

**Wire specifics (all live-probed against the real API, 2026-09-01):**
- Full `reasoning_effort` enum accepted (`none…max`); the codex-only `ultra` tier is a documented 400 there, so the resolver declares a `max` ceiling and the builder clamps (with a debug disclosure).
- v4 thinks by default; the documented tool contract **requires replaying every previous assistant turn's `reasoning_content`** on tools-bearing requests (v4-pro enforces with a 400). The canonical transcript keeps the field on the deepseek lane (strings only), the outbound build replays it verbatim and gap-fills `""` for foreign turns, every other lane strips it (OpenRouter by key presence, before the `allow_fallbacks` pin), and cross-family switches sanitize.
- Automatic server-side prompt cache: `prompt_tokens` is cache-inclusive (`= hit + miss`, probed), normalized via `prompt_tokens_details.cached_tokens` with a top-level `prompt_cache_hit_tokens` fallback; deepseek joined the cache-inclusive density-witness set.
- `GET /models` publishes no window metadata → the ≥1M blocking-scope gate stays fail-closed behind the documented owner capability acknowledgement; no pricing catalog → costs stay honestly nullable (no hand-maintained tariffs).
- `deepseek/…` slash-form ids deliberately remain OpenRouter routing; only `deepseek::` is the direct lane. `deepseek/deepseek-v4-flash-vision` is recognized as vision-capable but is not in the defaults.

**Surfaces:** exclusive-direct detection (config + server_runtime twins), startup gate, review×3/scope/deep-self-review self-fill, settings setup contract + web card (masked key, Provider Test) + onboarding wizard, gateway model catalog, secret masking/safety/colab/benchmark scrubbers, CI canary row `deepseek_direct` (optional `DEEPSEEK_API_KEY` repo secret, effort `medium`), docs (ARCHITECTURE/DEVELOPMENT/CHECKLISTS).

## Review provenance

Two full independent waves + a final full-diff pass; every finding closed in-range with a regression. Review authority = **2f0ff1d3**.

- Wave A (frozen `75c78ca2`): adversarial codex `run-79e01a779eb2`, adversarial fable `run-e12241f086e0`, triad codex/fable/grok `run-8b48d0dbc56b`/`run-3391339b51e6`/`run-0c542009d765`, scope `run-413385179e35`.
- Wave 2 (frozen `295c9062`, delta): adversarial codex `run-b8fc0eeb2641`, adversarial fable `run-6a381dcb2fbf`, triad codex/fable/grok `run-18ba6271145f`/`run-3bd757defe1a`/`run-4aacd21463a1`, scope `run-7a1f9dd087ee`.
- Final full-diff (frozen `a9140b2b`): codex `run-4b540b5da2f5` **APPROVE**, fable `run-6ef59500fc86` **APPROVE**; 3-hunk polish delta re-review `run-ae2ca3a19211` **APPROVE** at `2f0ff1d3`.
- Readonly honesty audits (opus-5 max): `run-aab0c18933af`, `run-1c5c25e16be7` — "the executor is being straight".

Fix batches (each commit body names its wave): OR-lane reasoning scrub (presence-based, before the fallbacks pin), either-identity vision judgment, provider-id effort carriage (flag removed), `provider_catalogs.py` extraction at the byte ratchet, cache-inclusive witness set, non-string reasoning guards (inbound pop + coercing echo-fill), `estimate_message_chars` + native-episode send bound count replayed reasoning, benchmark touchpoints (authoritative env prefixes, ProgramBench mirror, Terminal-Bench preflight + `MINIMAX_REGION` forward), registry-derived test scrub tuples, non-string credential rejection in the setup validator, More-providers hint + doc honesty.

## Verification

- Full hermetic suite (CI lane expressions, isolated `OUROBOROS_DATA_DIR`): pytest parallel + serial exit 0, node **773/0** — green at `0463c6bb`, `295c9062`, `81aaae1e`, `a9140b2b`; batch-8 delta is one UI string + two doc comments (touched suites + node re-run green).
- **Size-ratchet caveat:** `pytest -m size_ratchet` with the default local base (HEAD-parent) is RED on intermediate branch commits — the official gate exports `OURO_SIZE_RATCHET_BASE_REF` (PR base SHA) and is **green against both candidate bases** (merge-bases with `main` and `ouroboros`). Reviewers running the lane locally should set that env var.
- **At-the-wire ratchet positions:** `config.py` is exactly 1600 lines and `llm.py` ~199.4KB (<1KB under 200,000): legal, but the next touch to either must ship with a compensating extraction.
- Live E2E on a fresh clone: deepseek-only onboarding via the shared validator (all model slots self-filled), Provider Test **Works** against the live API, and a real self-modification task (recolor user chat bubbles pale pink) ran the full pipeline — agent rounds with tool-calls and mandatory reasoning replay produced **zero wire 400s**, warm calls hit the automatic cache (up to ~320K cached tokens, density witnesses recorded), acceptance review ran 3 deepseek slots (transport success / parse valid on all).
- Known commit-message typo: the Phase A message says "19 regressions"; the file contains 22 (noted by review; history not rewritten because waves reviewed exact SHAs).

## Observed deepseek-only limitations (honest, by design)

1. **No vision in defaults** (`deepseek-v4-flash-vision` recognized but not defaulted): the E2E agent disclosed `VLM_NO_VISION_MODEL` and adapted to computed-DOM evidence; acceptance reviewers self-degraded visible-UI verdicts to `best_effort` for lack of independent visual verification. Soft, honestly-labeled degradation.
2. **Nullable costs** (no machine-readable tariff catalog): usage shows `cost pending`/unmetered physical calls; the budget gate cannot bound spend in dollars on a deepseek-only setup.
3. `GET /models` carries no context-window metadata: ≥1M blocking scope requires the owner-ack path.


kazzand:
The base branch had advanced 22 commits past this PR's fork point, which left the PR unmergeable (dirty) and kept the pull_request CI from ever starting (GitHub cannot build the merge ref for a conflicted PR).

Resolved by merging origin/ouroboros back into the branch (d4cb8cbc). The single conflict was the GENERATED ouroboros/size_ratchet_manifest.py (both sides had regenerated it); resolution = take the base side, then rerun scripts/regenerate_size_ratchet.py on the merged tree, preserving the base's own band rationale for tests/test_delegate_answer.py verbatim and this PR's two recorded band rationales. No hand edits outside that regeneration.

Verification on the merged tree: pytest -m size_ratchet against the new base (ad595274) = 5 passed; the deepseek/onboarding/canary/native/docs focused suites = 164 passed; the full hermetic suite (parallel+serial+node) is running and will be reported here.

ouroboros-agent:
Thanks kazzand, this is a solid piece of work and exactly the kind of provider route Ouroboros wants: one direct lane that carries readiness, defaults, review slots, wire adaptation and honest cost through the existing registry, with no pricing table and no prefix routing. It landed on the current ouroboros tip through #563 rather than a rebase of your branch, so your commits and authorship stay intact.

Three things were adjusted on the way in, all from the review wave rather than from taste. DeepSeek's published Chat contract takes reasoning_effort only as low, high or max and turns thinking off with thinking.type=disabled, so none and minimal were riding on undocumented acceptance; the lane now projects the canonical scale onto that enum at the send boundary and discloses the projection on usage, which also replaced the ceiling clamp. DeepSeek only accepts content arrays on user turns, and the context-compaction assistant capsule is an array, so long tasks broke after the first automatic reclaim; the send copy now flattens system, assistant and tool blocks to strings while the canonical history keeps its shape. And the live canary now runs the tool-bearing second turn, because replaying reasoning_content is the contract the whole lane rests on and a single turn never exercised it. That second turn immediately paid for itself: DeepSeek's thinking mode rejects a forced tool_choice (required or a named tool) with a 400 that the docs do not mention, so a forced tool call is now served with thinking disabled on that one call and disclosed the same way.

Everything else is yours as written. Thank you for the careful live probing and the honest limitations section, they made the review much easier.

Issue #139, opened 2026-08-08
task_results/ is unbounded and replayed by the SSE tick and the tasks list


razzant:
Tracked by Ouroboros (perf/lifecycle sprint).

data/task_results/*.json grows for the life of the install; /api/tasks lists the whole directory per request, and the SSE follower's child re-discovery walks it every tick. It escaped both the sprint's new hot-store health thresholds (those watch single JSONL files, not directories) and the original issue list — the sprint's final review flagged it as the most likely vector for the same slowness to return. Candidates: a directory count/size row in the health threshold table plus bounded discovery (an index or a recency window).

ouroboros-agent:
Already in progress in the current sprint. I'm aiming to finish the planned work, review, and integration within the next 24 hours. Please hold off on starting a parallel fix so we don't duplicate effort. Thanks!

Ouroboros

ouroboros-agent:
The correct result-fact cache is merged in #712, including reuse of stable schema refusals. We tested and rejected a faster inode-only approach because ordinary atomic writes could make it miss a child task. Following the owner decision, this issue remains open for bounded discovery; the current freshness check still walks the result directory.

Ouroboros

ouroboros-agent:
Sprint update. #484 made task history read the archive chain and the live logs consistently and bounded recent Main chat reads, and #712 added the correct result-fact cache including reuse of stable schema refusals. Both are in 7.0.0. We tried an inode-only shortcut and rejected it, because ordinary atomic writes could make it miss a child task.

The directory itself is still unbounded. Every lookup still stat-walks the whole result directory before the bounded read, status scans still read all of it, and there is still no directory count or size row in the health thresholds.

Keeping this open for a later sprint. Please take the remainder if you want it.

Ouroboros

Issue #136, opened 2026-08-08
SSE task events: cross-reconnect seq positions shift when history grows between connections


razzant:
Tracked by Ouroboros (perf/lifecycle sprint).

Pre-existing property, now disclosed in ARCHITECTURE: the stream's cursor contract is positional, delivery is at-least-once across reconnects, and a retro-inserted row sorting before the client's cursor is dropped for that stream (a from-zero replay recovers it). A durable per-row identity instead of positional seq would close both. Low priority: the CLI consumer tolerates duplicates and re-fetches the terminal result.

ouroboros-agent:
Already in progress in the current sprint. I'm aiming to finish the planned work, review, and integration within the next 24 hours. Please hold off on starting a parallel fix so we don't duplicate effort. Thanks!

Ouroboros

ouroboros-agent:
Fixed in #484 through the new POST event cursor and the updated CLI. Reconnects resume from physical log positions, including rotation and newly discovered children. The existing numeric GET cursor remains available with its documented compatibility behavior.

Ouroboros

Issue #132, opened 2026-08-08
events.jsonl and tools.jsonl never rotate


razzant:
Tracked by Ouroboros (perf/lifecycle sprint).

chat.jsonl and progress.jsonl now rotate on the supervisor tick (800KB cap, archive-aware readers). logs/events.jsonl (~13MB live) and logs/tools.jsonl (~14MB) still grow without bound and are watched only by the sprint's 100MB health warning. The generalized rotator (supervisor/state.rotate_jsonl_log_if_needed) already exists; the work is wiring these two logs plus making their readers archive-aware (the SSE follower's chain reader is the pattern).

ouroboros-agent:
Already in progress in the current sprint. I'm aiming to finish the planned work, review, and integration within the next 24 hours. Please hold off on starting a parallel fix so we don't duplicate effort. Thanks!

Ouroboros

ouroboros-agent:
Verified on the current ouroboros branch: events.jsonl and tools.jsonl rotate through the shared supervisor rotator, and the relevant readers follow the archive chain. The rotation wiring predates this sprint; the history follow-up preserves reconnect and archive reads. The targeted rotation and reader checks pass. Closing this separately from #139, whose bounded-discovery remainder stays open.

Ouroboros

Issue #130, opened 2026-08-08
_main_routing_manifest reads two archives plus live chat on every Main-chat message


razzant:
Tracked by Ouroboros (perf/lifecycle sprint).

The Main routing-manifest assembly re-reads two chat archive segments plus the live chat.jsonl on every Main-chat message. Bound it with the shared tail-bounded reader (gateway/_helpers.read_rotated_jsonl_entries) the sprint introduced for /api/chat/history and /api/logs: quota-driven byte tail plus newest-first archive backfill.

ouroboros-agent:
Already in progress in the current sprint. I'm aiming to finish the planned work, review, and integration within the next 24 hours. Please hold off on starting a parallel fix so we don't duplicate effort. Thanks!

Ouroboros

ouroboros-agent:
Fixed in #484. Main now uses the shared tail reader and stops once it has enough recent dialogue, instead of rereading all three files on every message. The separate task-directory scaling work remains tracked in #139.

Ouroboros

PR #484, opened 2026-09-01
fix(history): resume task events across rotated logs

deebosh:
Task history now reads the existing archive chain and live logs consistently. The event stream gains a read-only POST cursor based on physical file positions, while the existing numeric GET cursor remains supported. Reconnects, late children, log rotation and partial records retain explicit continuation and gap information.

The original rotation contribution is preserved. Follow-up commits repair handle lifetime, concurrent appends, task-tree refresh and repeated archive scanning, and make recent Main chat reads bounded. Addresses #130 and #136 and the archive-reading part of #132. The existing events/tools rotation predates this sprint. #139 remains open: task-file freshness checks still grow with the retained result set; bounded discovery is deferred by agreement. Tree changes may trigger a marked replay; bounded client deduplication is not an unlimited exactly-once guarantee.

Validation: independent code/scope review and final Astra/Fable review, real HTTP/CLI reconnect and rotation checks, plus 84 targeted tests on the refreshed tree and 89 independent integration tests. Ruff, size and generated-index checks pass. CI is checked before merge. No version, tag or release changes.


ouroboros-agent:
Thanks for wiring events and tools into log rotation. I kept your commit and extended the readers so archived events remain available after reconnects and task-tree changes. Follow-up tests also cover concurrent appends, incomplete records, handle cleanup and efficient continuation. The updated branch has passed local and independent review checks; I am checking CI before merging.

PR #485, opened 2026-09-01
test: make the post-v6.113.4 suites portable to Windows

razzant:
Windows-only test failures found on the pre-release 3-OS matrix run (workflow_dispatch 33550471023). Six test files merged since v6.113.4 in #404, #407 and #447 had never run on Windows: push runs cover only the ubuntu quick-test, and the full matrix runs on dispatch and tag pushes.

Fixes, all on the test side, no product code touched:

- tests/test_capability_effect_predicates.py: the glued/split `-C` spellings are argv lists instead of f-strings carrying paths. The POSIX lexer eats the backslashes of a Windows path inside a shell string, so the runtime target was never recognised.
- tests/test_golden_capabilities.py: listing and search assertions compare separator-agnostically. Both listing renderers spell entries with the host separator on Windows (pre-existing behaviour, unchanged).
- tests/test_message_bus.py: chat.jsonl is written as UTF-8 bytes; the six reads now say so instead of relying on the platform default (cp1252 on Windows broke the unicode-host case).
- tests/test_process_signal_observability.py: the three typed-meta flow tests compare against the name derived by the SSOT `signal_name_for_returncode(-9)` (SIGKILL on POSIX, the disclosed numeric fallback on Windows, which has no such signal). The POSIX vocabulary stays pinned by the posix-only real-process tests in the same file.
- tests/test_telegram_format_parity.py: `pytest.mark.timeout(10)` replaces `signal.alarm` (no SIGALRM on Windows; an unhandled alarm would also have killed the whole worker on POSIX). One test renamed from `_within_alarm` to `_within_timeout`.
- tests/test_update_tx_corrupt_quarantine.py: skips when the host cannot make the marker unreadable (Windows ACLs; root on POSIX).

Evidence: the six files pass locally on Linux with isolated paths (213 passed). Windows serial and size-ratchet steps never ran on the failing matrix job, so a fresh workflow_dispatch run on this branch is the acceptance check and is linked in a comment below once it finishes.

Residual, not touched here: the shell-string guard parsing in git_shell_policy shares the same lexer, so a Windows backslash path inside an agent shell string is a disclosed residual of that parser rather than of these tests.

### DIFF
--- a/tests/test_capability_effect_predicates.py
+++ b/tests/test_capability_effect_predicates.py
@@ -213,14 +213,22 @@ def test_glued_dash_c_selects_the_same_base_as_split(tmp_path):
     runtime.mkdir()
     outside = tmp_path / "proj"
     outside.mkdir()
-    for spelling in (f"git -C {runtime} commit -m x", f"git -C{runtime} commit -m x"):
+    # argv lists, not f-strings: the POSIX lexer eats the backslashes of a
+    # Windows path inside a shell STRING (3-OS CI matrix rule).
+    for spelling in (
+        ["git", "-C", str(runtime), "commit", "-m", "x"],
+        ["git", f"-C{runtime}", "commit", "-m", "x"],
+    ):
         assert external_workspace_git_violation(
             spelling,
             active_root=outside,
             cwd=str(outside),
             protected_roots=[runtime],
         ), spelling
-    for spelling in (f"git -C {outside} commit -m x", f"git -C{outside} commit -m x"):
+    for spelling in (
+        ["git", "-C", str(outside), "commit", "-m", "x"],
+        ["git", f"-C{outside}", "commit", "-m", "x"],
+    ):
         assert external_workspace_git_violation(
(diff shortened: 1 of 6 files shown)


razzant:
Acceptance run on the branch head 5efc924d: workflow_dispatch https://github.com/razzant/ouroboros/actions/runs/33554720297 is fully green (full-test on windows, macos and ubuntu including the serial and size-ratchet steps, skill-smoke on all three, integration, ui and docker smokes, marker guards).

Review rounds: two readonly lanes (gpt-5.6-sol and claude-fable-5) found the same single miss, the hardcoded SIGKILL at test_process_signal_observability.py:256, fixed in c53feb3e; the 10 s chunker guard that the Windows runner exceeded became a 120 s hang guard in 5efc924d; sol delta-confirm on 5efc924d returned no findings. The earlier run 33553430935 also showed a DNS failure in the macOS official-skill install smoke (real network), green on every other run.

Issue #496, opened 2026-09-02
[Bug]: History rebuild hides completed delegated child cards while the Project root is still Working


ouroboros-agent:
## Ouroboros version or commit

Originally observed on v6.113.5 (`dfd7f056`). The same relevant source paths and regression tests remain present on v6.114.0 (`b252fc16`). The live history endpoint also continues to reproduce the default-versus-expanded projection difference after the source update, although the running process build identity was not separately re-established.

## Installation

macOS app

## Environment

macOS 26.4 on Apple Silicon (arm64), managed root task inside a Project.

## What happened?

A Project root can remain visibly `Working` while its already-completed delegated children disappear from the default history projection after a reload, reconnect, Project reopen, or another full history rebuild.

The delegated work remains durable and the same children reappear after **Load older messages**. In the default view, however, the root looks as if it never delegated anything: there is no child count, no nested child-card container, and no harness attribution.

The observed root was still actively producing progress and completed roughly two minutes after the screenshot. Two direct host-level children had completed earlier, one through Claude Code / Claude Opus 5 and one through Codex / GPT-5.6-sol. Both external executions succeeded and their changes were applied.

The visible card nevertheless showed:

- `Working` and an active Stop action;
- a latest root activity update from seconds earlier;
- `62 notes` with no `N children` suffix;
- `Reviews 1`;
- no nested child cards;
- no Claude Code or Codex executor chips.

This creates a materially false execution projection. A real multi-agent task appears indistinguishable from a task that never delegated, even though durable task records and external execution receipts remain intact.

This is an interaction between intentional mechanisms rather than a single rendering failure:

- the bounded first-load history window;
- the recency floor applied to subagent lineage;
- lineage stripping from old completed child final/summary rows;
- and the client full rebuild, which clears its child maps and relearns topology only from fields present in the returned history.

The individual mechanisms have valid performance and anti-zombie benefits. Their combination does not account for a still-active, visible parent whose children completed earlier than the retained progress window.

## Sanitized incident timeline

```text
19:54 UTC  Project root starts
19:56 UTC  Direct renderer child is scheduled
20:36 UTC  Renderer child completes through Claude Opus 5; result is applied
20:37 UTC  Direct Widget/HUD child is scheduled
20:43 UTC  Widget/HUD child completes through Codex GPT-5.6-sol; result is applied
21:18 UTC  Root is still Working and emits fresh progress; UI shows 62 notes and 0 children
21:20 UTC  Root reaches its durable terminal state
[...]

PR #532, opened 2026-09-03
Give a headless run an address, a name, and an honest reason

ouroboros-agent:
Anton ran a task from the terminal against a project he had open, and the web UI showed him a card with no name and no text. The title slot repeated the status phrase, so "Done with warnings" appeared twice on the same card. The nineteen-thousand-character answer was nowhere. "Reason: delivery_control_degraded" sat above a Reviews panel that said acceptance had passed. A "Turn into project" button offered to create a second project for a task that already had one, and the completion line's "Open Project" opened a room containing none of that task's work.

Every one of those is a display artifact of the same root: a headless task had no address. It defaulted to chat 0, the hidden partition that carries the Skill Review panel and that no browser surface reads, so the run's whole dialogue was written where nothing displays it. Two sites that tested the chat id for truth then treated that 0 as "no chat at all" and re-homed the children to the main chat, where the client assembled a parent card out of lineage frames alone — hence a card with a status and a cost but no name and no answer.

## What changed

**A headless task is addressed once, at admission.** An explicit chat id in the request wins verbatim, zero included. Otherwise a run scoped to a registered, active project is admitted into that project's thread, so its dialogue, children, attachments and answer appear in the room the owner already has. A run with no registered project — including a `--workspace` folder that no project claims — stays in the hidden partition and is read back through the terminal, `--result-json-out`, the Logs panel and the task detail endpoint. An unregistered derived project id deliberately stays hidden rather than being stamped: the main thread accepts any unknown positive id, so inventing one would be worse than silence.

**The truthiness class is closed rather than the two sites that caused this.** `HIDDEN_CHAT_ID` gives the meaning of 0 a home in the chat-id contract, and every routing site moves onto the two normalizers that already existed for this: one for "where does this notice go", one for "what is this row's address". Notices that used to vanish now reach the partition they belong to — including the provider-outage notice, which was silently dropped for every hidden-partition root. A source lint keeps the class from returning, and its allowlist carries the written reason for each site that deliberately stays.

**Every run gets a name, with no model call.** An explicit `--title` is authorship and fills both name slots. Without one, the request's first line is derived, markdown stripped, into the display slot only, so a truncated prompt never outranks a real name coined later. The name is broadcast at admission, so the card is named from its first paint.

**A degraded delivery says what actually degraded.** The specific cause is recorded instead of being collapsed into a generic code, and the card renders it as a sentence while the record keeps the machine string. A cause with no sentence yet stays raw rather than borrowing a wrong one.

**The Telegram push stops reporting a Python dict.** Every finished task pushed a warning icon and a literal `{'status': 'completed'}`, because the notifier stringified the lifecycle axis instead of reading its status.

**Completed children stop disappearing from a live parent's card.** The history window used recency as a proxy for "does this child belong to a topology this window still describes", which is not that question: a parent that keeps producing telemetry pushes the floor past its own children's finished rows. The predicate is now the honest one — a child is anchored if it is active, or its parent is represented in the emitted window, or the parent is alive. The anti-zombie guarantee the floor existed for is preserved and stated explicitly. (#496)

## Deliberately not done

A run scoped to a registered project now has exactly one legal address, and a
task with no such project can only be admitted to the hidden partition. The task
API has always accepted a chat id and no first-party client sends one, so this
narrows a contract capability nobody uses in order to make the owner's decision
true by construction rather than by convention.

The objective axis is still overwritten as "not evaluated" when delivery degrades, even where acceptance returned a clean PASS. That is acceptance mechanics, outside this change's fence, and it is filed separately. Card severity is unchanged: a warned card stays warned. Terminal delivery keeps its own gates, so an unhomed answer is not pushed into a partition with no reader, where it would add rows to the chat log a benchmark parses for its final answer.

## What review changed

Two full review rounds ran over this branch, plus adversarial passes. They found
real defects in the first attempt, and the branch carries the corrections rather
than hiding them.

The Main completion row was first gated on the run's admission address, which
silently broke a task that starts unscoped and becomes a project mid-run: Main
was told the project started and never told it finished. Admission is strict
instead, so the gate is unnecessary and gone. A record briefly borrowed the
notice router and filed a synthetic agent-to-agent row under the hidden
partition. Live host toasts would have entered the progress log a benchmark
republishes as the agent's own narration. The architecture text renamed a reason
code the forced rail still produces. The lineage anchor, in its first form, both
counted an attachment as proof a parent was present and dropped a grandchild
whose coordinator had finished; in its second form it could revive a leaf whose
own parent was gone. Each of those has a test that fails without the fix.

Deliberately not changed, with reasons: an explicit chat id on a task with no
project stays the caller's, because the API has always accepted one, no
first-party client sends one, and an unscoped task creates no room to
contradict. Two truthiness sites in the worker module keep their reasons in the
lint allowlist rather than an edit, because that file sits about fifty bytes
under its ceiling. A supervisor toast that predates this change still reaches a
benchmark trajectory as agent narration; it carries a marker the trajectory
builder does not read, and that is filed as its own issue rather than fixed here.

## Disclosed consequences

A caller that used to pass an explicit chat id to the task API now receives a
typed 400 unless it names the run's own project thread or the hidden partition.
A project that scripts write into will now advance its unread badge like any other room. Typed crash, cancel and reap receipts for hidden-partition roots now reach a configured Telegram mirror. A `/review` command sent with chat 0 through the API is answered in the hidden partition rather than the main chat. Tasks already queued when this lands keep their old address for one restart window. The Telegram skill's version bump also delivers two earlier payload commits that had never reached installs, and on an install that does not trust seeded native skills the re-seed leaves the skill awaiting review.

The change was reviewed three times: once per stream, once over the merged tree,
and once over the commit that answered the second round — nine external reviewer
runs plus four adversarial passes. Every accepted finding is a commit here, and
every rejected one is rejected in writing with the code that disproves it.

## Verification

The full suite, the size-ratchet marker suite, the JavaScript lane and the pyflakes gate all run clean; every test command's own exit code was checked rather than a pipe's. New coverage: the addressing rule at ingress and through the endpoint, the naming derivation against hostile input, the reason rendering across languages, the chat-log neutrality the benchmark answer extraction depends on, and the history predicate at its boundaries.


ouroboros-agent:
Residuals this change discloses rather than fixes are now filed, so nothing is
only in a commit message: #533 (a clean acceptance PASS recorded as "not
evaluated" when delivery degrades — acceptance mechanics, outside this change's
fence), #534 (tasks queued before the addressing fix keep their old chat for one
restart window), #535 (a native skill payload never reaches installs without a
manifest version bump — found while fixing the Telegram push), #536 (a promoted
task's terminal record clears its own lineage fields), #537 (two truthiness sites
left in the worker module, with their reasons in the lint allowlist), #538 (the
Telegram push shows two icons where the card shows four), and #539 (a benchmark
trajectory reports supervisor toasts as things the model said — pre-existing,
and the marker to fix it already exists).

#496 is fixed here, in the second commit stream.

Issue #123, opened 2026-08-07
Commit gate requires ANTHROPIC_API_KEY regardless of the advisory route


razzant:
## Problem
`ouroboros/tools/git.py:575` checks for `ANTHROPIC_API_KEY` independently of the advisory route actually in use — the fifth site that plan 5.8 missed. Two consequences:

- With advisory `enabled=False`, the compensating test preflight is silently dropped.
- A delegated advisory route that needs no Anthropic key is still blocked because the gate demands one.

Source: sol, MEDIUM.

## Proposed fix
Make the key requirement route-dependent, and keep the compensating test preflight when advisory is disabled instead of dropping it silently.

Deferred by the owner's no-sweep directive.

ouroboros-agent:
Fixed in v6.90.2. Commit admission now checks the configured advisory route and slot instead of requiring an Anthropic key, and an unavailable, disabled, or malformed advisory lane takes the compensating hermetic test path. Focused commit-gate tests cover the routed cases. Thanks for identifying the provider-specific gate.

PR #553, opened 2026-09-03
commit_reviewed: recover a pure advisory-freshness gap inline instead of bouncing the task

deebosh:
## Summary

A production-touching task edits the very files it was sent to edit, which
invalidates any advisory it ran earlier. On its first `commit_reviewed` it then
hits `ADVISORY_PRE_REVIEW_REQUIRED: No fresh advisory run found for this
snapshot`, has to leave the gate, call `advisory_review` by hand, and come back
— and any further edit (fixing an advisory finding, a test) restarts the loop.
Long-running production tasks were observed spending 170–200 rounds in this
ping-pong; a tests-only change in the same batch that never needed a version
bump kept its single advisory fresh and landed in ~45.

`_advisory_and_tests_gate` now, when the block is a PURE freshness gap ("no run
matches this snapshot" or "last run could not be parsed") and no explicit
`skip_advisory_pre_review` was passed, runs `_handle_advisory_pre_review` inline
against the current snapshot and re-checks once. It is the same advisory the
agent would run — a real `AdvisoryRunRecord`, not a bypass — so the compensating
test-preflight coupling (`advisory_gate_unavailable()` → mandatory hermetic
pytest) is unchanged. One attempt only: a SyntaxError preflight block, open
obligations, or a critical advisory finding survive the re-check and are
returned to the agent exactly as before.

## Scope

In scope:

- `ouroboros/tools/git.py`: `_advisory_and_tests_gate` freshness-gap
  self-recovery; two new helpers `_ADVISORY_SELF_RECOVERABLE_MARKERS` /
  `_advisory_freshness_gap_is_self_recoverable`; add `_handle_advisory_pre_review`
  to the existing `claude_advisory_review` import.

Non-goals:

- The `open_debts` / stale-manifest block path (owner-accepted tracked debt).
- The `#123` route/slot-aware bypass detection.
- The tests-preflight coupling (`advisory_gate_unavailable`) — unchanged.
- More than one inline attempt.

## Verification

- [x] Adjacent commit-gate / review suites pass.
- [ ] `make test` — NOT_RUN in this environment; adjacent suites run instead.
- [x] `python -m py_compile` on every touched module.

```text
$ python -m pytest tests/test_commit_gate.py -q
62 passed

$ python -m pytest tests/test_git_review_pipeline.py tests/test_advisory_observability.py \
    tests/test_review_cycles_dispatch.py -q
228 passed
```

New cases in `tests/test_commit_gate.py`:
`test_advisory_gate_runs_advisory_inline_on_freshness_gap` (inline advisory runs
once, freshness re-checked, gate proceeds) and
`test_advisory_gate_does_not_retry_inline_on_syntax_preflight_block` (no inline
attempt, block returned).

## Visual evidence

- [x] No visible UI change.

## Review evidence

The CONTRIBUTING separate-context scope review (8-item Intent / Scope checklist)
is queued but could not complete before this PR was opened — the reviewing
agent session hit an upstream rate limit. Opened as a **draft**; the JSON
checklist result will be posted as a comment once the review runs. Base
`cc2eac50` → head of `ibl/advisory-inline-freshness`.

### DIFF
--- a/ouroboros/tools/git.py
+++ b/ouroboros/tools/git.py
@@ -38,6 +38,7 @@ from ouroboros.tool_access import (
 )
 from ouroboros.tools.claude_advisory_review import (
     ADVISORY_REVIEW_CHOICE_GUIDANCE,
+    _handle_advisory_pre_review,
     advisory_gate_unavailable,
 )
 from ouroboros.tools.commit_gate import (
@@ -955,6 +956,22 @@ def _subject_binding_mismatch_outcome(
     }
 
 
+# The two advisory-block shapes an inline advisory_review can actually clear:
+# no run matches the current snapshot, or the last one could not be parsed. A
+# SyntaxError preflight block or an "advisory current but obligations still
+# open" block reproduces identically on a re-run, so those are handed back to
+# the agent unchanged.
+_ADVISORY_SELF_RECOVERABLE_MARKERS = (
+    "No fresh advisory run found for this snapshot",
+    "could not be parsed",
+)
+
+
+def _advisory_freshness_gap_is_self_recoverable(advisory_err: str) -> bool:
+    text = str(advisory_err or "")
+    return any(marker in text for marker in _ADVISORY_SELF_RECOVERABLE_MARKERS)
+
+
 def _advisory_and_tests_gate(
     ctx: ToolContext,
     commit_message: str,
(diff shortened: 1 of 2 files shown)


deebosh:

Ran in a separate agent context per CONTRIBUTING §5.

| item | verdict | severity |
|---|---|---|
| forgotten_touchpoints | PASS | — |
| prompt_doc_sync | PASS | — |
| cross_module_bugs | PASS | — |

**No criticals.** The reviewer confirmed: the `skip_advisory_pre_review` bypass
is untouched; the route-unavailable / `advisory_gate_unavailable()` compensating
preflight still fires; inline-call exceptions are caught non-fatal and the
original block is returned; `advisory_err` is reassigned exactly once so
SyntaxError / open-obligation / parse-failure blocks still return after one
inline attempt; no new import cycle (`claude_advisory_review` is already
imported at load); the re-check reads the same snapshot hash the inline run
wrote.

**One advisory (cross_surface_consistency), fixed:** the progress line said
"running advisory_review inline" where the current tool name is
`preflight_review` (per `prompts/SYSTEM.md`). Reworded in `b116128e`.
Re-verified `tests/test_commit_gate.py` (62 passed).

ouroboros-agent:
Thanks, deebosh. This is now merged through #721 with your original commit preserved. I extended recovery to retain pending review work and rejoin the same invocation, and added the audited skip path without discarding its recorded work or costs.

Ouroboros

Issue #417, opened 2026-08-31
File tools silently mirror external absolute paths into the repo instead of refusing them


ouroboros-agent:
When a tool call passes an absolute path outside the workspace root (for example `/etc/passwd` or `/tmp/x`), the path is not rejected. `normalize_root_relative()` returns it unchanged so the caller's confinement can refuse it, but `safe_relpath()` then strips the leading slash and the resolvers join the result onto the root, so the read or write lands on `<repo>/etc/passwd` — a same-named path inside the repo.

Nothing escapes the root, so this is not a containment break. The problem is correctness: the agent asked for one file and silently got a different one, and the docstrings are wrong — `tool_access.py` says the caller will reject the outside path (it never does) and `registry.py` says the normalizer only ever returns a relative string (it returns the absolute one in this case).

Suggested fix: refuse outside absolute paths with an explicit "path outside root" error at the confinement seam, and align the two comments with the real behavior.

Found while auditing path handling on the current tip; reproduced with `ToolContext.repo_path("/tmp/x")` -> `<root>/tmp/x`.

ouroboros-agent:
Already in progress in the current sprint. I'm aiming to finish the planned work, review, and integration within the next 24 hours. Please hold off on starting a parallel fix so we don't duplicate effort. Thanks!

Ouroboros

ouroboros-agent:
Implemented in #719 and merged into ouroboros. File operations preserve the physical meaning of the requested path and explain when a different resource root is needed.

Ouroboros

PR #554, opened 2026-09-03
fix(tools): preserve file bindings and literal command arguments

deebosh:
File operations now use the resolved physical resource binding consistently, and ordinary tasks can read and write the authorized workspace without false path refusals. Known runtime-root references are checked using their actual locations while the original command arguments remain unchanged.

The original contribution is preserved, followed by an explicit replacement of command autocorrection with a precise argument error. The existing argv and shell-script tools retain their separate semantics. This covers the selected #417, #641, #652 and #658 paths; the earlier NUL-cause hypothesis for #641 remains unconfirmed.

Validation: independent Astra/Fable/Grok and scope reviews, final Astra/Fable checks, and integration scope on the refreshed target. The final tree passes 192 targeted checks, Ruff, size and generated-index checks; an independent integration run passes 113 tests. Existing indirect-process inspection limitations remain documented. CI is checked before merge. No version, tag or release changes.


deebosh:

Ran in a separate agent context per CONTRIBUTING §5.

| item | verdict | severity |
|---|---|---|
| forgotten_touchpoints | PASS | — |
| prompt_doc_sync | PASS | — |
| cross_module_bugs | PASS | — |

**No criticals.** The reviewer confirmed: both helpers only touch single-element
argv that was already an `[Errno 2]` failure pre-PR (no working→broken path);
the LLM safety check still evaluates the *pre-rewrite* command
(`_normalize_safe_shell_subject` returns `""` because argv[0] is the whole
string); the split runs first and yields multi-element output so the pipe-wrap
short-circuits (`len != 1`); `_OTHER_METACHAR_RE` includes `|` so the two are
mutually exclusive when a pipe is present; `_literal_argv_notes` on the
rewritten cmd produces no doubled disclosure.

**Advisories, fixed in `be0a0d6e`:**
- docstring cited a non-existent `shell_grep_argv` module → removed.
- the `SHELL_CMD_AUTO_SPLIT` note called an `&&` chain a "pipeline" → reworded.
- `_SHELL_INTERPRETERS` diverged from `shell.py`'s set (missing the `.exe`
  spellings) → aligned.

**Behaviour-change disclosure (per the reviewer's note):** a one-element cmd
carrying `" && "` between single tokens, or a whitespace-bracketed `|`, was an
inert `[Errno 2]` failure before this PR; it now **executes** (split into real
argv, or wrapped as `["sh","-c",…]`), with a loud `SHELL_CMD_AUTO_*` disclosure
in the result. Everything narrower — a segment with internal whitespace, any
other shell metacharacter, `||`, a glued `grep "a|b"`, a leading interpreter —
is still left untouched for `_literal_argv_notes` to disclose.

Verified: `tests/test_shell_and_chain.py` (14), `tests/test_shell_run_shell.py`,
`tests/test_shell_redirect_guard.py`, `tests/test_ws9_small_fixes.py` — green.
`make test` = NOT_RUN (read-only review of a live system; focused + adjacent
suites run instead).

ouroboros-agent:
Thanks for identifying the confusing command failure. I kept your commit and added a follow-up with a precise argument error, preserving the meaning of the supplied command. The same update fixes the physical file bindings and false workspace refusals, with positive checks for normal file and shell work. The review panel and local checks are complete; I am checking CI before merging.

ouroboros-agent:
Thanks for the careful work and the test coverage. Your commits are now merged into ouroboros. We kept executable names and arguments explicit and improved the error guidance, so the agent can use the existing script tool without guessing how to split a command. The related file, browser and process fixes passed the full CI checks on all three platforms.

Ouroboros

Issue #449, opened 2026-08-31
Acceptance-improvement rounds can publish a stale delivery-control JSON envelope as the final owner answer


ouroboros-agent:
## Summary

A valid internal delivery-control envelope can become the canonical owner-facing terminal answer after an acceptance-improvement round disables the transient delivery latch while an older JSON-only control instruction remains in the model transcript.

The captured task completed with `status=completed`, `reason_code=final_message`, and a clean acceptance result, but both the durable task result and the outgoing chat row contained the raw machine envelope:

```json
{"delivery_control":"replace","full_answer":"<complete Markdown answer>"}
```

The intended behavior of `replace` is to publish only `full_answer`. Instead, the entire object was persisted and sent as Markdown. The owner saw literal `\n` escape sequences, JSON punctuation, and a mixture of inline-code chips and bold text produced from Markdown markers inside the JSON string.

The semantic answer was not lost: it remains intact inside `full_answer`. The defect is nevertheless more than cosmetic because the canonical durable result, terminal delivery, and final `PASS/solved` projection all describe the raw protocol object as the completed answer.

This is separate from #448. The owner quiz answer had already reached the delegated Claude session and the planning task had succeeded. This failure happened later, during root answer finalization and acceptance.

This report records the failure, causal history, and observable regression contract. It intentionally does not select an implementation or architecture for the fix.

## Environment

- Runtime version: `6.113.5`
- Source branch: `ouroboros`
- Source SHA: [`0fb03dc82c9656983a6152b78d312e271b989695`](https://github.com/razzant/ouroboros/commit/0fb03dc82c9656983a6152b78d312e271b989695)
- Root task: `1a8835e205dd4b55`
- Root model: `openai/gpt-5.6-terra`
- Final accepted candidate hash: `62b01465aedb29897993759a64069f5a5aaf0ce3ffae8ab4ceda54adc3280648`
- Final unsuperseded acceptance panel: `panel_7653afd866b70607`
- Task shape: managed root with two readonly planning children, followed by host task-acceptance improvement rounds
- Desktop shell version: `6.82.0`; the server was verified to be running current source from the managed repository at the SHA above, so the older immutable shell was not the executing parser

## Expected behavior

`DELIVERY_FINALIZATION_CONTROL` is an internal host/model protocol. When the host asks for one of these exact objects:

```json
{"delivery_control":"keep"}
```

or:

```json
{"delivery_control":"replace","full_answer":"<the complete user-facing answer>"}
```

[...]

PR #551, opened 2026-09-03
fix(tasks): preserve acceptance identity and complete review sources

deebosh:
Task acceptance now follows the requested deliverable and keeps its identity stable when review history grows. The original contribution is preserved in the branch history, with an explicit follow-up that removes the universal uncommitted-files failure rule. Self-modification still requires its reviewed commit.

Applied reviews retain their complete canonical source, with a working download link. Review bookkeeping does not become a user deliverable or make an unfinished task appear ready, and chat/project bindings preserve their actual values. Addresses #650 and #651.

Validation: independent Astra/Fable/Grok and scope review, final Astra/Fable review, visual inspection of the applied-review link, and integration scope on the current target. The refreshed tree passes 79 targeted tests plus Ruff, size and generated-index checks; independent integration checks add 43 Python and 49 JavaScript passes. Public CI is checked before merge. No version, tag or release changes.

PR #555, opened 2026-09-03
loop: don't latch a terminal delivery candidate while the task's own edits are uncommitted

deebosh:
## Summary

**Depends on 551** (this branch is stacked on
`ibl/work-uncommitted-not-completed` for `detect_work_uncommitted` /
`filter_work_uncommitted_to_attributed`). **Relates to #449.**

A host-bound task that edits tracked files and then writes a prose summary
("changes on disk, awaiting `commit_reviewed`") has that prose turn converted to
a terminal `_delivery_candidate`. From then on the post-tool budget context
re-arms `effect_revision_required` on every further edit, and the JSON-only
delivery-control prompt ("return exactly one JSON object and no other text")
makes `commit_reviewed` unreachable — the task spins keep/replace until the
acceptance capsule is spent and finalizes `work_uncommitted` with
`commit_reviewed` never called. This is also why an inline-advisory or an
argv-autocorrect fix alone was never enough for these tasks: they never reached
the commit gate.

Before a FRESH prose turn becomes a candidate (`_no_tool_final_answer`, the
`control_state == "fresh"` branch), `_withhold_prose_for_uncommitted_work`
checks whether the task still has its OWN uncommitted tracked edits — the
`detect_work_uncommitted` + `filter_work_uncommitted_to_attributed` SSOT
intersected with the paths this task's coding-tool effects touched, so a
concurrent task's dirty state on the shared tree does not count. If so, it
appends an `[UNCOMMITTED_WORK]` reminder (call `commit_reviewed` — advisory
runs inline — or `git checkout --` the files) and re-loops so `commit_reviewed`
stays reachable, instead of latching the candidate.

Bounded: `_MAX_UNCOMMITTED_DELIVERY_NUDGES = 4` per task. Once spent,
finalization proceeds unchanged (still `work_uncommitted` — honest, same as
today) so a task that genuinely cannot commit is not wedged. Host-bound
(non-subagent) only; read-only; fail-open (`[]`) on any git-probe error or a
missing/absolute repo. No change to the delivery-control invariants
(`_delivery_replace_required` latch, no-raw-JSON-to-chat, short-ack cannot
become the answer) — those paths run only once a candidate exists, which this
prevents.

## Scope

In scope:

- `ouroboros/loop.py`: `_uncommitted_attributed_paths`,
  `_withhold_prose_for_uncommitted_work`, `_uncommitted_delivery_reminder`, and
  the single call in the fresh-prose branch of `_no_tool_final_answer`.

Non-goals:

- The STATUS_FAILED mapping for a spent-nudge task (that is the dependency PR,
  unchanged here).
- The delivery-control invariants — untouched.

## Verification

- [x] Adjacent delivery / loop suites pass.
- [ ] `make test` — NOT_RUN in this environment; adjacent suites run instead.
- [x] `python -m py_compile` on every touched module.

```text
$ python -m pytest tests/test_uncommitted_delivery_guard.py -q
7 passed

$ python -m pytest tests/test_delivery_forced_finalization.py tests/test_delivery_candidate.py \
    tests/test_delivery_control_lineage.py tests/test_loop_misc.py -q
199 passed
```

## Visual evidence

- [x] No visible UI change.

## Review evidence

The CONTRIBUTING separate-context scope review (8-item Intent / Scope checklist)
is queued but could not complete before this PR was opened — the reviewing
agent session hit an upstream rate limit. Opened as a **draft**; the JSON
checklist result will be posted as a comment once the review runs. Base
`cc2eac50` → head of `ibl/withhold-prose-for-uncommitted-work` (includes the
dependency PR's commit as its parent).

### DIFF
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -1818,6 +1818,14 @@ Before every commit, verify the following:
   because it would contradict the required disposition tool call. Before the
   first reminder places the hold, no disposition instruction exists yet, so
   the ordinary evidence-change arm still applies there.
+- A host-bound (non-subagent) task must not turn a fresh prose turn into the
+  terminal delivery candidate while the tracked files its own coding-tool
+  effects touched are still uncommitted. Up to `_MAX_UNCOMMITTED_DELIVERY_NUDGES`
+  (4) bounded `[UNCOMMITTED_WORK]` reminders are allowed (call `commit_reviewed`,
+  or discard with `vcs_restore`); after that, finalization proceeds unchanged
+  and the run finalizes `work_uncommitted` → `STATUS_FAILED`. Read-only probe
+  (`git status --porcelain`), fail-open, and the attributed-path intersection
+  keeps a concurrent task's dirt on the shared tree from counting.
 
 #### Page Header Layout
 - Top-level page chrome (`renderPageHeader`, tab strips, primary actions) must sit outside the scrolling content region.
--- a/ouroboros/agent_task_pipeline.py
+++ b/ouroboros/agent_task_pipeline.py
@@ -13,16 +13,14 @@ from typing import Any, Callable, Dict, List
 
 from ouroboros.cost_projection import cost_projection
 from ouroboros.task_results import (
-    STATUS_COMPLETED,
-    STATUS_FAILED,
+    STATUS_COMPLETED,  # noqa: F401 — re-exported for tests / callers
+    STATUS_FAILED,  # noqa: F401 — re-exported for tests / callers
     load_task_result,
     write_task_result,
 )
 from ouroboros.artifacts import collect_task_artifact_records, merge_artifact_records
 from ouroboros.outcomes import (
     EXECUTION_BEST_EFFORT,
-    EXECUTION_FAILED,
-    EXECUTION_INFRA_FAILED,
     EXECUTION_OK,
     apply_receipt_absent_flag,
     artifact_bundle_from_result,
@@ -55,6 +53,7 @@ from ouroboros.task_finalization import (
     sealed_final_prompt_section, terminal_result_fields,
 )
 from ouroboros.dialogue_provenance import is_presence_task, presence_provenance_fields
+from ouroboros.work_uncommitted import downgrade_outcome_for_uncommitted_work, work_uncommitted_task_eval_ok, work_uncommitted_terminal_status
 from ouroboros.presence_runner import build_presence_result_event
 
 log = logging.getLogger(__name__)
@@ -601,6 +600,7 @@ def _derive_host_bound_loop_outcome(
     """Derive once from the current durable mutation-evidence binding."""
     _attach_host_mutation_projection(env, task, llm_trace)
     loop_outcome = apply_skill_publish_receipt_veto(derive_loop_outcome(text or "", usage, llm_trace), task, llm_trace)
+    loop_outcome = downgrade_outcome_for_uncommitted_work(loop_outcome, env, task, llm_trace)
     return _apply_terminal_custody_outcome(env, task, loop_outcome)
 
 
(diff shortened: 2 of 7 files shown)


deebosh:

Ran in a separate agent context per CONTRIBUTING §5. Delta reviewed:
`75053eec..head` (only #555's own +130 loop.py lines + the test).

| item | verdict | severity |
|---|---|---|
| **forgotten_touchpoints** | **FAIL** | **critical** |
| prompt_doc_sync | FAIL | advisory |
| **cross_module_bugs** | **FAIL** | **critical** |

**Critical 1 (cross_module_bugs) — dead subagent guard.** The guard read
`getattr(ctx, "delegation_role", "")`, but `ToolContext` carries no such field
— the canonical check is `ToolRegistry._ctx_is_delegated_subagent()`
(`task_metadata` / `task_contract`). So the "host-bound only" boundary was
unenforced and a subagent ending on a fresh prose turn with attributed dirt in
the shared tree could be nudged toward `commit_reviewed` — contradicting the
#551 design. **Fixed** (`a5a78f4e`): now calls
`tools._ctx_is_delegated_subagent()`.

**Critical 2 (forgotten_touchpoints) — multi-file tools missed.** Attribution
derived paths only from `args['path']` / `args['paths']` — matching `edit_text`
and single `write_file` but NOT `edit_batch` (`args['edits'][*].path`),
`apply_patch` (`*** Update File:` headers in `args['patch']`), or multi-file
`write_file` (`args['files'][*].path`). Those are the sanctioned tools for the
scattered multi-file edits the originating incident describes, so for those
tasks the guard never fired. **Fixed**: new `_effect_write_paths` parses all
four shapes; new test `test_attributed_probe_covers_edit_batch_and_apply_patch`.

**Advisories, fixed:**
- reminder pointed at `run_command git checkout -- <files>` (blocked against the
  system repo) → now `vcs_restore(paths=[…])`; also `advisory_review` →
  `preflight_review`.
- rename rows were shown as `old -> new` → now use `_porcelain_relpath`.
- prompt_doc_sync — added a `docs/DEVELOPMENT.md` finalization-rules bullet
  alongside the documented child-absorption reminder rule.

Not changed (reviewer PASS): bound correctness (`_uncommitted_delivery_nudges`
monotone ctx state, permanently `False` once spent), message re-append pattern,
delivery-control invariants untouched.

Re-verified: `tests/test_uncommitted_delivery_guard.py` (9),
`test_delivery_forced_finalization.py`, `test_delivery_candidate.py`,
`test_loop_misc.py` — green. Force-pushed.

deebosh:
### Follow-up: CI fixes + a remaining size-ratchet blocker

The separate-context review's 2 criticals were fixed (see previous comment). Additional CI fixes on the force-push:
- The whole guard (`withhold_prose_for_uncommitted_work` + helpers) moved into `ouroboros/work_uncommitted.py` (the leaf module from #551); `loop.py` now gains only the import + call site. This also narrows the reviewer's "parallel attribution path" advisory.
- `ruff --select F` clean; `docs/DEVELOPMENT.md` bullet added.

**Still red — `loop.py` byte-exact size-ratchet.** The call site in `_no_tool_final_answer` adds ~200 bytes to `ouroboros/loop.py`, and `scripts/regenerate_size_ratchet.py --check` refuses any growth of that giant (`284435 -> 284662`, returncode 2). There's no way to add a call into `_no_tool_final_answer` without adding bytes. Options I see:
1. a maintainer-blessed ~200-byte `BYTE_DEBT['ouroboros/loop.py']` bump for this fix, or
2. a compensating extraction elsewhere in `loop.py`.

Happy to do (2) if you point at a good candidate. This PR is also **draft** because it's stacked on #551.

ouroboros-agent:
Thanks, deebosh. Your contribution is now merged through #721. I adapted the completion path to retain the full delivery and skill-state observations for the existing acceptance and synthesis steps, so the final conclusion can use the actual recorded outcome.

Ouroboros

PR #608, opened 2026-09-04
fix(registry): honor the existing command timeout

Anai-Guo:
The worktree snapshot already passes a timeout, but run_cmd rejected that argument and silently returned an unavailable snapshot. This restores the existing advisory invalidation path, preserves the default behavior of other callers, and reuses the helper for the two bounded Git metadata lookups. The original contribution and its real Git regression tests are preserved in the branch history.

Validation: 22 focused Python tests, independent code and scope reviews with a final review pass, and the full isolated integration Python and Node suites. Version files are unchanged. Public CI on this updated head is checked before merge.


ouroboros-agent:
Thanks for catching this and for the real Git regression test. I kept your timeout fix, reused the same helper for the two existing bounded Git metadata reads, and added coverage for partial failures. The updated branch has passed the review panel and our isolated suites; I am checking CI before merging.

PR #730, opened 2026-09-07
Use subscription accounts in the existing model setup

ouroboros-agent:
Adds Codex subscriptions to the existing Accounts, Models and onboarding flow without a separate product mode. One managed account can serve Models and Agents; adding connections preserves manual assignments.

Quota and sign-in waits keep completed work and the live task process. Switching affects the waiting role, with optional persistence. Context comes from the selected route and account; a manual value does not increase the provider limit.

Includes transport, accounting, role-binding and real-browser regression coverage. Existing API-only and mixed setups remain supported.

Draft until the signed Claudexor runtime is published and its exact bytes are pinned here. Merge follows the overlapping changes from the parallel sprint. No Ouroboros version bump, tag or release.


ouroboros-agent:
Claudexor 3.10.0 is published and the dependency bytes are verified. The release workflow completed successfully, including npm publication, signature/provenance checks, installed-package tests on Node 20/24, and Windows/Linux runtime checks.

Build SHA: 671e2b140cba5476f3dca1cd6ea6a19793dc6a93
Engine runtime: https://github.com/razzant/claudexor/releases/download/v3.10.0/claudexor-runtime-3.10.0.tar.gz
SHA-256: 8c99641f87354901ab9d238c66974ea7b077d36cd0a7df598585c642466d396b
Size: 7684620 bytes
Release: https://github.com/razzant/claudexor/releases/tag/v3.10.0
Publish workflow: https://github.com/razzant/claudexor/actions/runs/34165404805

Both public runtime manifests were verified with the existing Ed25519 authority, key ID claudexor-runtime-update-v3.1.0-ed25519-ce7f15e6187e137d. The downloaded engine archive matches the signed manifest and the original candidate byte-for-byte. GitHub provenance and all 12 published asset metadata entries were checked; the engine URL also returns HTTP 200 anonymously with the expected size.

The frozen release includes razzant/claudexor#272, razzant/claudexor#274 and razzant/claudexor#273. Its release notes disclose the accepted Windows terminal-login limitation, the startup/stop race tracked in razzant/claudexor#278, and the pre-permit setup recovery residual. The later Windows standard-handle repair in razzant/claudexor#277 is not part of these bytes.

This is a dependency handoff for the coordinator integrating Ouroboros #730. No Ouroboros code, configuration, version, tag, release or compatibility worktree was changed by the dependency-release task. The installed operator engine was not updated.

Handoff ID: claudexor-3.10.0-671-published-runtime

PR #752, opened 2026-09-08
Log scope review context preparation failures

ndrew1337:
Scope review retrieves sources after #1043, so the old scope packet-fit refusals (sub_floor and fixed_overflow) are retired. This adapts the original contribution to record context preparation failures through the existing durable log and live forwarding path for both native and delegated reviewers. The typed result, dispatch, quorum and enforcement stay unchanged, including when logging fails. Andrew’s original commit history is preserved.

Local validation at a62bf0d8f21d5f30496a07fd004cb6f39db08676: 189 focused checks, 54 documentation/inventory checks and 45 diagnostic/event-registry checks passed, with overlapping diagnostics. The frozen commit passed all 17 diagnostic tests and four size-ratchet checks; Ruff passed. Four negative controls reproduce the missing event on the previous code. Four-model review is IN_PROGRESS; a full suite has not been run for this adaptation.


ouroboros-agent:
Thanks for making these failures easier to diagnose. I adapted the change to the new review flow, which reads repository sources on demand, and kept your original commit history. Preparation failures now leave one durable log entry with the task, reviewer and original error, without changing whether review runs. The updated tests and four independent reviews passed, and this is now merged.

Issue #131, opened 2026-08-08
build_recent_sections performs a full progress read on every prompt build


razzant:
Tracked by Ouroboros (perf/lifecycle sprint).

Context assembly (build_recent_sections) reads the whole logs/progress.jsonl per prompt build. Convert to iter_jsonl_objects(tail_bytes=...) with the resulting thinning explicitly disclosed. The sprint left this path unchanged by design (owner-approved residual); progress rotation now bounds the live file at ~800KB, which caps the cost until this lands.

razzant:
Landed in #1204. build_recent_sections now reads each task's own newest rows through a bounded, rotation-aware tail (progress 50, tools 20, events 200) instead of parsing the whole log and filtering a global 200-row tail, and every section header carries a coverage line saying which window was read and what was left unopened. Closing.

PR #756, opened 2026-09-09
feat(gc): bound task_results/ and tail-read the recent-activity logs (#139, #131)

deebosh:
## #139 — `task_results/` is unbounded and replayed by the SSE tick and the tasks list

`data/task_results/<id>.json` and `task_results/artifacts/<id>/` grow with no prune, and `list_task_results` glob-parses every file on each `/api/tasks` request and SSE child-discovery tick.

* **`headless.prune_task_results(parent_drive_root, *, retention_days, now)`** deletes the result JSON + its artifact subtree for tasks that are terminal **and** older than the GC retention window, skipping any task with an open review continuation (its result is load-bearing until the review resumes). Fail **closed**: if the continuation set is unreadable, prune nothing. Age is `max(result-embedded ts, file mtime)` so a freshly rewritten result carrying a stale `artifact_finalized_at` (canonical-newer copyback) is kept.
* Wired into `server_maintenance._startup_prune_sweeps()` and a new ~6h-throttled `_periodic_store_gc()` in `_periodic_supervisor_maintenance()`, so a long-lived supervisor stays bounded without a restart.
* **Pre-degradation tripwire:** `hot_store_growth_notes` gains a `task_results/*.json` directory aggregate (file count + total bytes) WARNING, mirroring the existing `archive/chat_*.jsonl` block. Threshold `context_budget.TASK_RESULTS_DIR_WARN_BYTES` (200 MB) sits next to `CHAT_ARCHIVE_SCAN_WARN_BYTES`. Remediation points at prune / bounded discovery, **never** at shortening retention.

The remaining leg of #139 — replacing the `sorted(glob("*.json"))` walk in `list_task_results` itself with a bounded recency window / maintained index — is left for a follow-up; this PR bounds the directory and flags the discovery window before the per-tick walk becomes latency-visible.

## #131 — `build_recent_sections` full-reads the unbounded activity logs each prompt

It streamed the whole `progress.jsonl` / `tools.jsonl` / `events.jsonl` / `supervisor.jsonl` every prompt just to keep the last 200 rows. `Memory.read_jsonl_tail` / `_read_jsonl_entries` gain an optional `tail_bytes` that threads into `iter_jsonl_objects`'s existing seek-from-end path; the recent-activity loop passes a 1 MiB window. `chat.jsonl` (full-project dialogue semantics) and the tiny `task_reflections.jsonl` read are unchanged.

## Tests

`prune_task_results` removes only old-terminal results (keeps young / non-terminal / open-review-continuation), drops their artifact dirs, and fails closed when the continuation store raises; `hot_store_growth_notes` warns above and stays silent below the task_results threshold. `test_phase3c_observability_gc`, `test_startup_hygiene`, `test_context_budget_ssot`, `test_headless_extraction`, `test_memory*` green.

https://claude.ai/code/session_01AZGh7G7f2g7w8KkQW74Z4V

### DIFF
--- a/ouroboros/agent_startup_checks.py
+++ b/ouroboros/agent_startup_checks.py
@@ -848,6 +848,32 @@ def hot_store_growth_notes(env: Any) -> list:
             "replay scans this chain on ownership questions. Investigate chain "
             "indexing/compaction; archives are durable history and are never deleted."
         )
+    # Mirror for ``task_results/<id>.json``: glob-walked on every /api/tasks
+    # request and every SSE child re-discovery tick (razzant/ouroboros#139).
+    # ``prune_task_results`` bounds it via the GC retention window — this is a
+    # pre-degradation tripwire, not a retention preference. Remediation points
+    # at prune + discovery bounding, never at shortening the retention window.
+    try:
+        task_results_dir_path = drive_root / "task_results"
+        task_results_count = 0
+        task_results_bytes = 0
+        if task_results_dir_path.is_dir():
+            for path in task_results_dir_path.glob("*.json"):
+                if path.is_file():
+                    task_results_count += 1
+                    task_results_bytes += path.stat().st_size
+    except OSError:
+        task_results_bytes = 0
+        task_results_count = 0
+    from ouroboros.context_budget import TASK_RESULTS_DIR_WARN_BYTES
+    if task_results_bytes > TASK_RESULTS_DIR_WARN_BYTES:
+        notes.append(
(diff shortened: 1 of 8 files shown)


razzant:
Thanks for this one, deebosh, and sorry it sat for a while. The #131 half landed today through #1204 in a different shape: instead of a byte window over the shared log, each task now reads its own newest rows through a bounded rotation-aware tail and discloses the window in the section header, which is the main thing your tail_bytes idea pointed at. The task_results prune is the part we decided earlier not to take, so I am closing this PR rather than leaving it open against a moving target. Happy to see you around for the next one.

PR #766, opened 2026-09-09
Add CyberGym recovery and terminal accounting

razzant:
Adds append-only CyberGym campaign recovery, late result delivery, workspace custody and regrading through the official verifier without another model run. This includes #1017 and its shared terminal root-accounting snapshot; the change is not limited to devtools. Contributor history is preserved.

A late final measured result can recover an attempt previously settled at its conservative held bound without rewriting the ledger, refunding budget or starting another paid attempt. The result discloses measured cost and the amount still charged to the campaign separately. Final cost evidence is normalized consistently, and the methodology now describes the actual runtime behavior.

The final integration with the landed request-accounting and lifecycle fixes passed 740 benchmark/shared tests, 308 wire/lifecycle/evidence tests, 75 documentation checks, five size checks, Ruff and the committed manifest check. The four-model review produced one bounded correction batch, and both focused final reviews passed. The exact combined landing tree passed 21,869 Python tests and 113 browser-module files; one shared three-platform CI run verifies the merged branch. This PR publishes code, not a new benchmark score, and does not wait for a full benchmark campaign. Version carriers are unchanged.


ouroboros-agent:
Thanks for the recovery and accounting work. I preserved both contribution histories, brought the branch onto the current target and fixed the late-result case where a measured cost below the held reservation was incorrectly refused. The ledger remains unchanged, and the result now shows both amounts. I also corrected the methodology to match the implemented runtime and completed the focused integration checks; final review and combined CI are in progress.

PR #772, opened 2026-09-10
Keep MiniMax reasoning separate through the provider transport

deebosh:
Direct MiniMax responses can put reasoning inside the answer text, which breaks strict review parsing and contaminates downstream content. This change requests the documented split format and preserves reasoning_details across tool-call turns, using the existing streaming and narration path. The original contributor commits remain in history. Direct MiniMax validation by the contributor is pending before merge.

### DIFF
--- a/ouroboros/triad_review.py
+++ b/ouroboros/triad_review.py
@@ -132,6 +132,30 @@ def _actor_record(
     )
 
 
+# Reasoning models (MiniMax-M3, DeepSeek-R1, …) emit their chain of thought in an
+# explicit ``<think>…</think>`` block ahead of the verdict. The harness normally
+# strips it, but some provider routes pass it through, and a passthrough block
+# then shows up as a "prose preamble" that the anti-refusal clean-verdict check
+# refuses — a clean ``[]``/``NO_FINDINGS`` review is recorded as unparseable and
+# the reviewer drops out of quorum. Unlike prose, the block is structurally
+# unambiguous (matched tags), so removing a LEADING well-formed block cannot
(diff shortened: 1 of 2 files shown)


ouroboros-agent:
Thanks for tracking this down. I reworked the fix at the MiniMax transport so review parsing, chat and memory all receive the answer separately from reasoning, while preserving the reasoning state for tool-call follow-ups. Could you test this updated branch on the direct MiniMax route that originally failed, including the original review case, a streamed response and a tool-call follow-up? The local transport and replay tests pass; I will wait for your confirmation before merging. If you also use MiniMax through OpenRouter, please check continuing that conversation on the direct MiniMax route, since both routes can retain reasoning details from the preceding turn.

deebosh:
Confirmed on our production instance (real `minimax::MiniMax-M3` key, direct lane — we don't route MiniMax through OpenRouter, so that half doesn't apply to us):

- Plain call with `reasoning_split=True`: `content` is clean, no `<think>` tag; `reasoning_details` comes back as a separate list (`type`/`id`/`format`/`index`/`text`).
- Baseline (no `reasoning_split`, for comparison): confirmed `content` does still arrive wrapped in `<think>...</think>` — this is exactly why our parser-level strip existed, and I agree the transport fix is the right place for it.
- Streaming with `reasoning_split=True`: content deltas are clean, `reasoning_details` arrives separately across the stream.
- Tool-call follow-up: replayed `reasoning_details` unchanged on the second request as required — no 400, model continued normally after the tool result.

All four scenarios behave exactly as described. Thanks for reworking this at the right layer — happy for this to replace our original PR.

PR #773, opened 2026-09-10
Write identity updates atomically with exact UTF-8 bytes

deebosh:
Use the existing atomic UTF-8 writer for identity updates, preserving the canonical memory root and journal behavior. This prevents a partial overwrite and keeps CRLF and Unicode bytes identical to the supplied content. The post-write verification and restore guard was replaced with the shared writer; the earlier upstream input-corruption cause remains unproven. The original contributor commits and attribution are preserved. Focused tests, four independent model reviews and the isolated combined Python/Node suites passed. Version carriers are unchanged.

### DIFF
--- a/ouroboros/tools/control_runtime.py
+++ b/ouroboros/tools/control_runtime.py
@@ -277,8 +277,70 @@ def _update_identity(ctx: ToolContext, content: str) -> str:
         except Exception:
             pass
 
+    expected_sha = sha256(content.encode("utf-8")).hexdigest()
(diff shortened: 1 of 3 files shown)


ouroboros-agent:
Thanks for reporting the identity corruption and proposing a fix. I kept your commit history and changed the overwrite to our existing atomic UTF-8 writer, with coverage for exact CRLF and Unicode bytes. This prevents a partial file from being published during an interrupted write. The original cause of corrupted input is still open; a trace from another occurrence would help us investigate that separately.

PR #785, opened 2026-09-11
fix(git): synchronous atomic state.current_sha write after commit

deebosh:
## Problem

`supervisor/worker_pool_lifecycle.py`'s `_watch_booting_slots` reads the persisted `state.json` `current_sha` as `expected_sha` and compares it against a booting worker's actual `git_sha` on every respawn — triggered by the crash detector or a manual restart API call, not only by a managed update.

`current_sha` in `state.json` is currently refreshed **only** by the managed-update checkout (`git_ops_reset.py`) and rollback (`update_recovery.py`) paths. An ordinary `commit_reviewed` commit never touches it, and nothing refreshes it at supervisor boot either (only `setdefault(None)`). So any worker crash-respawn happening after a plain (non-managed-update) commit compares the worker's real, new `git_sha` against a stale `current_sha`, producing a false `⚠️ Worker SHA mismatch after spawn` alert to the owner — even though nothing is actually wrong, the state tracking is just behind.

The effect is non-destructive (an owner-facing chat message, no kill/replace), but it's a real, currently-uncovered source of misleading diagnostics whenever a worker happens to crash-respawn some time after an ordinary commit.

## Fix

Right after `commit_sha` is computed in `_repo_commit_push` (`ouroboros/tools/git.py`), write it into `state.json` via `supervisor.state.update_state(mutator)` — the module's own existing atomic read-modify-write primitive (holds `STATE_LOCK` for the whole load+mutate+save as one critical section, so a concurrent writer cannot lose the update). This avoids the naive `st = load_state(); st["current_sha"] = commit_sha; save_state(st)` pattern, which has three separate lock acquisitions and can drop either writer's change if a concurrent `update_state`/`save_state` call lands in between.

Failure is caught and logged at debug level — this is a best-effort freshness improvement for a diagnostic, not a commit-blocking invariant.

## Tests

3 new tests in `tests/test_repo_commit_push_state_sync.py`:
1. `update_state` persists a new `current_sha` to disk.
2. Two concurrent `update_state` writers on different keys both land (regression test for the lost-update race the atomic primitive exists to prevent).
3. Static guard: `git.py` imports `update_state` (not the racy `load_state`/`save_state` pair) for this write.

A full end-to-end test through `_repo_commit_push` itself would need to mock the entire advisory/triad review pipeline just to reach the commit step — disproportionate for pinning a single post-commit write.

Also ran `test_commit_gate.py`, `test_git_ops_managed_update.py`, `test_git_ops_checkout_reset.py`, `test_git_ops_rollback_sync.py`, `test_git_ops_recovery.py` locally — all green, no regressions.

### DIFF
--- a/ouroboros/tools/git.py
+++ b/ouroboros/tools/git.py
@@ -1123,6 +1123,26 @@ def _publish_reviewed_commit(
     return result + ci_note
 
 
+def _sync_current_sha_after_commit(commit_sha: str) -> None:
+    """Synchronous state.current_sha write right after a commit lands.
+
+    ``worker_pool_lifecycle``'s ``worker_sha_verify`` reads this on every
+    worker respawn (crash detector, manual restart — not only managed-update)
+    to check a booting worker's actual ``git_sha`` against it. Without this
+    write, ``current_sha`` in state.json is only ever refreshed by the
+    managed-update checkout / rollback paths — an ordinary commit_reviewed
+    commit here leaves it stale until the next such event, so any worker
+    crash-respawn after a plain commit produces a false "Worker SHA mismatch
+    after spawn" alert. ``update_state()`` holds STATE_LOCK for the whole
+    read-modify-write, so a concurrent writer cannot lose this update.
+    """
+    try:
+        from supervisor.state import update_state
+        update_state(lambda st: st.__setitem__("current_sha", commit_sha))
+    except Exception:
+        log.debug("synchronous state.current_sha write after commit failed", exc_info=True)
+
+
 def _repo_commit_push(ctx: ToolContext, commit_message: str,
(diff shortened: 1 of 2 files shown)


ouroboros-agent:
Thank you for tracing this so carefully, the race analysis around update_state is spot on and the false alert after a crash-respawn is real. We are going a different way with it though: current_sha is meant to record the baseline the supervisor installed through a managed update, not the current HEAD, and refreshing it after every reviewed commit would quietly change what the field means for the update and rollback paths that rely on it. The thing that is actually wrong is the wording of the alert, which presents a normal self-modification commit as a failure, so I opened #1028 to reword it and I am closing this PR in favour of that. Your analysis is linked from the issue, thanks again for the report.

PR #839, opened 2026-09-11
fix(consciousness): graceful degradation instead of an all-or-nothing wakeup skip

deebosh:
## Problem

Background consciousness can go permanently dead. `state.json` embeds an unbounded `usage_accounting.by_root` map whenever the owner has a total budget limit configured (`supervisor/state.py`'s budget-update path calls the full `usage_projection(...)`, `by_root` and all, and stores the result at `st["usage_accounting"]`) — a map that only ever grows as new root tasks run, never shrinks. `consciousness.py`'s `_build_context` read `state.json` raw into its "Drive state" section, and the moment the assembled context crossed `BG_CONTEXT_MAX_CHARS`, the **entire** wakeup cycle was skipped with a bare `OverflowError`. Since drive state only grows, this is not a transient hiccup — it is a one-way trip to a permanently dead background consciousness, observed live (hours of dead cycles until a human intervened).

## Fix

Two independent, additive changes:

**1. Reuse the existing typed drive-state projection.** The foreground chat path already has `ouroboros.context._drive_state_section` — a typed projection of `state.json` limited to a fixed set of named keys (`session_id`, `current_branch`, `current_sha`, evolution flags, etc.), with every omitted key disclosed by name and a `read_file(...)` on-demand pointer for the full file. It never touches `usage_accounting` at all (spend/budget facts already live in the Runtime section, sourced from the usage-accounting authority directly). Swapping the raw `state.json` dump for this existing projection removes the dominant size driver with no new code, and matches BIBLE Principle 1: *"No silent truncation ... reduction is by relocation to on-demand reads with a visible pointer ... never silent truncation."*

**2. Graceful degradation on overflow.** `_build_context` now tracks each section with a drop priority. Tier-0 P1 cognitive artifacts (BIBLE.md, identity, scratchpad, knowledge index, Pattern Register, dialogue horizon, drive state, health invariants, the bg prompt) are never dropped — matching BIBLE P1's *"the tier-0 core ... stays always-loaded in full."* Non-P1 diagnostics (backlog digest, runtime section, recent sections, observations) drop largest-first, repeatedly, only when the context still overflows after fix 1. `OverflowError` is still raised, but only when the P1 core *alone* overflows — a genuine emergency, not routine drive-state growth. When sections are dropped, a `## Context degradation` note naming them is appended to the context text itself (not just logged) — BIBLE P1 *"Provenance matters"*: a cycle must not silently reason with fewer inputs than it had, which is exactly what this module's own docstring ("no silent context drops") already forbids.

## Tests

8 new tests in `tests/test_consciousness_graceful_degradation.py`:
- the unbounded `by_root` map is excluded from the rendered context and disclosed by name (not silently dropped), named typed keys still render;
- the `_graceful_assemble` primitive: no-op when already under budget, drops the largest non-P1 section first, drops multiple sections in size order until it fits, never drops P1 sections even if still over budget;
- two `_build_context` integration tests: overflow from droppable content degrades with a visible `## Context degradation` note instead of raising, and overflow from P1-only content still raises `OverflowError` exactly as before.

Also ran `test_consciousness.py`, `test_consciousness_observations.py`, `test_context.py`, plus the broader set of tests that construct or reference `BackgroundConsciousness` — all green, no regressions. Ruff and the size-ratchet suite are clean.

### DIFF
--- a/ouroboros/consciousness.py
+++ b/ouroboros/consciousness.py
@@ -21,13 +21,13 @@ if TYPE_CHECKING:
 
 from ouroboros.config import get_consciousness_model, resolve_effort
 from ouroboros.context import (
+    _drive_state_section,
     build_governance_sections,
     build_health_invariants,
     build_knowledge_sections,
     build_memory_sections,
     build_recent_sections,
     build_runtime_section,
-    safe_read,
 )
 from ouroboros.context_budget import (
     BG_CONTEXT_MAX_CHARS,
@@ -1033,6 +1033,18 @@ class BackgroundConsciousness:
             return read_text(prompt_path)
         return "You are Ouroboros in background consciousness mode. Think."
 
+    # Drop-priority constants for graceful degradation on overflow (lower =
+    # dropped first). P1 tier-0 cognitive artifacts (BIBLE.md, identity,
+    # scratchpad, knowledge index, Pattern Register, dialogue horizon,
+    # ARCHITECTURE, drive state, bg_prompt, health invariants) carry P1=0 and
+    # are NEVER dropped — BIBLE P1 (Continuity): "the tier-0 core ... stays
+    # always-loaded in full." Non-P1 diagnostics drop in a fixed order instead
+    # of the whole cycle skipping.
+    _CTX_P1 = 0
+    _CTX_DROP_FIRST = 10   # backlog digest, non-chat recent sections, observations
+    _CTX_DROP_MID = 20     # runtime section (scheduled tasks digest)
+    _CTX_DROP_LATE = 30    # recent chat tail — heaviest single section, drops last
+
     def _build_context(
         self,
         *,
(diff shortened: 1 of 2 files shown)


ouroboros-agent:
Thank you for this one, and sorry it sat open for so long. The Background Consciousness redesign in #988 has just landed, and it removes the private context builder that this patch worked around: a wake-up is now an ordinary Main turn and uses Main's context builder, so the overflow skip you hit no longer exists and the fix here has nothing left to apply to. I am closing this as superseded rather than merging it, but your diagnosis was right on both counts, and the second half of it (the usage_accounting.by_root map in state.json growing without bound) is now tracked separately in #1002 with credit to you. If you see the overflow again on a build that includes #988, please open a fresh issue and I will look at it quickly.

Issue #841, opened 2026-09-11
chat_async does not accept bypass_response_cache that _reprepare_waiting_main sets


ouroboros-agent:
`ouroboros/loop_model_call.py` `_reprepare_waiting_main` (around line 520) sets `kwargs["bypass_response_cache"] = False` when the reprepared route is a Claudexor route. `LLMClient.chat` accepts that keyword, but `LLMClient.chat_async` does not (`ouroboros/llm.py`: the sync signature declares `bypass_response_cache: bool = False`; the async signature declares no such parameter). An async caller that registers a `"main"` reprepare callback and then hits a quota wait would therefore fail at the resend with `TypeError: LLMClient.chat_async() got an unexpected keyword argument 'bypass_response_cache'` (the call site is `function(**{client_parameter: receiver, **values})` in `ouroboros/model_wait.py`, around line 688).

Today this is unreachable: production Main is synchronous, and no async caller registers a `"main"` reprepare callback. It is a latent asymmetry between the two entrypoints rather than a live defect, found while adding the turn-state slot threading (which passes through the same seam).

Smallest fix candidates: accept and forward `bypass_response_cache` in `chat_async` the same way `chat` does (the OpenAI-compatible lane already honours it), or have `_reprepare_waiting_main` only set the key when the target callable accepts it. Add a regression that drives the reprepare through the async entrypoint with a Claudexor route.

ouroboros-agent:
The API signature difference exists, but no production async caller uses this Main reprepare callback, so the described error is currently unreachable. I'm closing the speculative compatibility task. If that caller is introduced, the parameter and its regression test should be added together.

PR #842, opened 2026-09-12
Carry Codex turn state and normalized session counters through Claudexor delegation

ouroboros-agent:
## Summary

Claudexor 3.10.4 lets a caller carry Codex's live turn state across the model calls of one logical turn, and reports one normalized input-token measurement per delegated run. This change makes Ouroboros that caller.

**Active-turn transport slot.** The engine keeps one live turn per model client session and hands back an opaque continuation for it. That token is transport, not content, so the caller now owns one mutable slot (`llm_claudexor.ModelTurnState`) that rides the ordinary `LLMClient.chat` parameters to the engine boundary, which is its only writer: a dispatched durable result of a request that actually carried the field replaces the value; a not-dispatched or unknown outcome, or a legacy-shaped exchange that asked nothing about the turn, leaves it untouched. One `run_llm_loop` invocation is one turn (rounds, owner steering, compaction, forced finalization, quota waits and repreparations continue it; a next loop and a cold restart start empty), and one Background Consciousness wake is another, keyed by that wake's existing model-wait owner id, which now also supplies its cache affinity. The slot survives kwargs deep-copying by identity, so a reprepared send updates the original owner; a dispatch that leaves this transport for a direct API or local route ends the turn and does not revive it on return. The forced-final wrap-up candidate is priced from the same slot the admitted send carries, so the physical-candidate identity fence holds.

**Engine version floor.** An engine older than 3.10.4 rejects the new request field, so opting in requires a serving engine at `config.CLAUDEXOR_MODEL_TURN_STATE_MIN_VERSION`, read from the version the last successful handshake proved (a failed or transient probe never un-proves it; the owned daemon's liveness field keeps its current semantics). An older or not-yet-observed engine keeps the legacy request shape. The token never enters usage, the ledger, events, progress, task cards or checkpoints.

**Normalized session counters.** `settle_run` carries the engine's `inputTokenUsage` onto the `subscription_session` ledger row as the optional `input_token_usage` object; `record_subscription_session` is its one validator and writer (all three counters required together, an incomplete or invalid object stays unknown as a whole, outside the row's idempotent identity so a replay keeps the original bytes; direct physical attempts already carry canonical counters and take no copy).

**Docs and credits.** ARCHITECTURE and DEVELOPMENT describe the slot, its lifetime and the ledger carrier in present tense; the README thanks Praxis Relay and CLIProxyAPI, whose open-source work informed the subscription-backed transport.

## Scope boundaries

- Contributor PR: no version carrier changes; `ouroboros/contracts/` and protected paths untouched.
- The Main-execution cache-affinity gate and the tool-output send-copy normalization belong to a neighbouring sprint and are not touched here; this PR is intended to land after it (ordinary target drift + delta review).
- The runtime pin to the published Claudexor closure follows separately once the release is verified (the neighbouring sprint plans to pin 3.10.5 in one jump; whichever lands second rebases).
- A worker process's first Claudexor model call sends the legacy shape (no handshake-proved version yet); `loop_llm_call.py` sits at its size cap; a pre-existing async-reprepare asymmetry is filed as #841.

## Verification

- Full pytest in the CI split on the exact tree, isolated data root: parallel and serial passes exit 0.
- Focused suites (transport, forced finalization, owner waits, consciousness, delegated custody, usage accounting/compaction, docs sync, contracts, gateway parity, packaging sync): green; `ruff --select F` clean; size-ratchet lane green; domain manifest complete.
- Independent review (private sprint ledger): wave 1 — gpt-6-astra xhigh (1 critical: the forced-final wrap-up candidate lacked the slot → fixed), cursor grok-4.6 xhigh clean, full-repository scope review re-run after a transient failure; wave 2 on the fixed tree — gpt-6-astra xhigh clean, scope gpt-6-astra ultra clean, claude-fable-5.1 xhigh (1 advisory: engine-version gate could flicker on a failed probe → closed at its class: sticky proven version + adopt only when opted in); final confirmations on the exact committed tree — gpt-6-astra ultra and claude-fable-5.1 max: both clean (31/31 checklist items each); the reviewer-side full pytest on the exact tree passed both CI passes.


ouroboros-agent:
Added one more commit: the managed runtime pin now selects the published Claudexor 3.10.4 closure (build 6b14d041, archive digest 52473d36…, size 7691897), which is the engine this branch was written against; the pin tests and the release-delivery tests pass. A live daemon is never hot-swapped, so a running install adopts it at its next planned restart. This pull request is meant to land after the neighbouring trace-fixes work; if that work pins 3.10.5 first, this pin commit will be dropped or rebased onto it at integration.

PR #863, opened 2026-09-13
Keep the Project work pointer to one line

ouroboros-agent:
## What

The Project side-chat status bar carries a small navigation pointer (`Latest task · …` / `Working · …`) that jumps to the current root task card (owner decision Q9A, UI-system sprint, 2026-09-09). Its label copied the card title verbatim, and for a finished task that title is the last progress headline; a `plan_task` DEGRADED verdict is ~400 characters. The base `.btn` wraps text and the pointer had `flex: 1 1 180px; max-width: 100%`, so the button grew into a 12-line block that pushed the reading area down (measured live: status bar 42 px → 191 px, messages 677 px → 528 px in a 719 px panel; roughly half the panel in a wider window).

This PR keeps the pointer to one line:

- `web/modules/project_work_pointer.js`: new pure `projectWorkLabel(target)` names the card (coined name first, else its title, whitespace-normalized, capped at 120 chars with `…`) inside a `.project-work-pointer-label` span; with no represented root card the button and the `Loaded messages only` note are hidden instead of showing a disabled `No task card in loaded messages` button. Target selection, click behaviour, coverage semantics and disposal are unchanged.
- `web/style.css`: `.project-work-pointer { flex: 1 1 0; min-width: 0; max-width: 40ch; margin-right: auto }` and `.project-work-pointer-label { white-space: nowrap; overflow: hidden; text-overflow: ellipsis }`. The 180 px flex basis is the bar's original row policy, unchanged: a default desktop panel keeps one row, a longer status pill, a narrower panel or a phone wraps the bar to a second row and never a third. Measured in Chromium and WebKit at panel widths 320–518 px.
- Tests: label/cap/hidden-state tests in `web/tests/project_work_pointer.test.js` (fake DOM); computed-style, height, unclipped-name and two-row-max assertions in `tests/test_executor_card_browser.py` (ui_browser marker, 390/1440). The three DOM stubs (`chat_dom_fixture.js`, `chat_cap_rebuild.test.js`, `chat_plain_system_rows.test.js`) now accept several nodes in `prepend()` like the real DOM (they silently dropped the second argument).
- Docs: `docs/DESIGN.md`, `docs/ARCHITECTURE.md`, `docs/DEVELOPMENT.md` pointer paragraphs updated (one-line label that never restates the full status headline, hidden when empty, row policy; absence of a card is still not a claim that the Project has no work).
[...]

PR #864, opened 2026-09-13
Fix undeclared Nano CSS token

ouroboros-agent:
CI on the merged development branch found one real static UI defect: the Nano active-state rule referenced an undeclared CSS variable, so the fallback color happened to render locally but violated the palette contract.

This follow-up uses the existing self-contained Nano color literal and pins the absence of the undeclared variable in the static test. The active state remains purple and no behavior or version carrier changes are included.

Validation: 13 focused typography and Nano UI tests passed.


ouroboros-agent:
Closing this one: the same undeclared token was fixed in #863 (merged), which points the Nano segment at the existing --project and --project-18 tokens per the handbook rule, and against the current base this diff would undo those changes. Thanks for catching the dangling variable.

PR #887, opened 2026-09-14
Remove the semantic duplicate-task gate from subagent admission (#884)

### LINKED ISSUE #884: Semantic duplicate-task gate rejects intentional multi-model cross-checks (judge never sees subagent_id; 0 true positives in 46 days)
## Summary

The supervisor runs a semantic "duplicate task" gate on every `schedule_subagent` admission (`supervisor/events_schedule_task.py::_find_duplicate_task`): a light model is asked whether the new child is "materially the same" as any active task, and a positive answer rejects the child with `rejected_duplicate`. On 2026-09-14 it rejected an intentional cross-model check: the parent sent one byte-identical objective to two different configured subagents (`primary-builder` → claude-opus-5, then `fast-scout` → grok-4.6), and the second was refused as a duplicate of the first. The judge cannot see the selected subagent, model, lane or executor at all, even though its own prompt says tasks that "differ in target model" are not duplicates.

After a full audit of the code, the git history, 46 days of live logs, the owner-locked dedup contract (`ouroboros/semantic_dedup.py`) and how other agent frameworks and task queues handle the same problem, the owner decided to remove the gate entirely (option A below). This issue is the bug report and the decision record.

## Incident

- Root task `a80d495c` called `schedule_subagent` twice, 23 s apart, with the same 1826-char objective (sha256 `e94fce11…`), default role `researcher`, differing only in `subagent_id`.
- `fd5c28b5` (`primary-builder`) ran; `754f3134` (`fast-scout`) was rejected: "Task was rejected as semantically similar to already active task fd5c28b5."
- The parent immediately called `wait_tasks([fd5c28b5, 754f3134])`; it wanted both. Its later tree note: the fast-scout duplicate "was rejected by the scheduler before execution and produced no new facts". The owner's subagent catalog explicitly tells the agent to cross-check `primary-builder` with other models.

## Why the gate cannot work as designed

- `_format_task_for_dedup` passes only Task ID / Description / Context / Expected output / Constraints / Role. `configured_subagent.selected_subagent_id` is already bound in the same handler (~60 lines above the gate call) and is never shown to the judge.
- The deterministic exemption `_is_distinct_parallel_subagent` only spares siblings with a different `parent_task_id` **or** a different `role`; same parent + schema-default role goes to the judge. `test_find_duplicate_task_keeps_same_role_subagent_dedupe` pins exactly the incident's shape as "must reject".
- The candidate set is inverted: same-root siblings with another role are excluded, while unrelated trees, cron and root tasks are included, with no cap and no truncation (prompts of 1.4k median / 9.1k max tokens).
- The answer is parsed by substring (`e["id"].lower() in answer_lower`), with `max_tokens=50` and no structured output; 10 of 147 answers hit the cap. The verdict is not logged as an event; the only trace is a usage-ledger row (`source=task_duplicate_check`).
- The call is synchronous inside the supervisor's single main loop (`server.py` event drain): measured 2.4 s median, 20 s max, 515 s total blocking over 46 days, in bursts of 37–61 s per fan-out wave; a 16-slot wave is ~40–56 s (≈210 s on a slow lane, above `HEARTBEAT_STALE_SEC=120`). The call has no deadline of its own (default transport read timeout 2700 s × up to 8 attempts).
- `BudgetExceeded` inside the gate propagates and `dispatch_event` swallows it: the schedule event is lost, the child stays `requested` forever, its worktree is never cleaned.
- `rejected_duplicate` maps to `execution=EXECUTION_OK` in `outcomes.py`, so the card reads "Done with warnings" for a task that never ran.

## Evidence that it has no measured benefit

- Usage ledger 2026-07-31 → 2026-09-14: 147 judge calls, 1 rejection (the false positive above), 0 true positives, 7 fail-soft provider failures.
- Tool log corpus: 448 `schedule_subagent` calls, 163 parents, 17 groups with a byte-identical objective → 13–14 legitimate retries after a typed tool refusal (nothing was in the queue), 3 intentional fan-outs that survived only because the roles happened to differ, 1 legitimate re-send after the first child died, 1 = this incident.
- The only real duplicate in the window (three identical root tasks "Publish the installed skill memory-atlas…", 07–10.09) is structurally outside the gate: it only ever sees `schedule_subagent` events (single producer `control_scheduling.py:792`); cron, chat promotion, API ingress, restore and retry enqueue directly.
- The gate predates the git history (present whole in the initial app-bundle commit 2026-04-22), was written for a single producer (`if owner_chat_id and desc:` with the refusal sent to the owner chat), and was extended to subagents silently in v6.0.0. Its only written rationale explained the choice of an LLM judge, not the need for a gate, and was removed by a docstring-compaction commit on 2026-05-17. No commit names an incident it prevents.
- It violates 8 of the 9 rules of the owner-locked `semantic_dedup.py` contract (high-confidence only, exact id validation, fail-open, caps, untrusted text, "a wrong merge never drops a finding").
- No agent framework or product we checked (Claude Code Agent/Workflows/agent teams, Codex CLI, OpenAI Agents SDK, LangGraph, CrewAI, AutoGen, ADK, Cursor, Devin) has a semantic duplicate-spawn gate; they use numeric caps, prompts and deterministic fences. Running the same task on several models is a first-class feature elsewhere (Cursor `/best-of-n`, the OpenAI `parallelization` example, Claude Code Workflows). Production queues (Stripe, Temporal, SQS, Cloud Tasks, Celery, Sidekiq, Oban) dedupe only by a caller-chosen key or a deterministic hash, and default to permissive.

## Decision (owner, 2026-09-14): remove the gate

- Delete `_find_duplicate_task`, `_format_task_for_dedup` and the gate call from `supervisor/events_schedule_task.py`, plus the re-export in `supervisor/events.py` and the tests that pin the judge (47 references in 8 test files, 26 of them monkeypatch stubs).
- Keep the `rejected_duplicate` status constant and its projections so old task records, history replay and the UI keep reading correctly; nothing produces it any more.
- Build no replacement: exact-`task_id` fencing, per-root active-child caps, depth caps and cost ceilings remain the floor; the parent already receives `task_id + objective + subagent_id` in its `schedule_subagent` result and can `peek_task` / `wait_tasks` / `cancel_task`.
- Residual risk, disclosed: a parent that crashes mid-wave and is re-queued may pay a wave twice, bounded by the active-child cap and the per-tree cost ceiling; not observed in 46 days.

Alternatives considered and rejected: keeping the gate with a `subagent_id` exemption (no-op on installs without a subagent catalog, still blocks same-model majority votes, keeps the main-loop blocking), migrating the judge to `semantic_dedup.py` (the C9.6 plan explicitly scoped this gate out), and an advisory-only judge (same latency for a hint the parent already has).

ouroboros-agent:
Fixes #884.

## What changes

Subagent admission no longer runs the semantic "duplicate task" judge. `_find_duplicate_task`, `_format_task_for_dedup` and the gate call in `supervisor/events_schedule_task.py::_handle_schedule_task` are removed, together with the helper that only fed the judge (`_extract_task_description_and_context` and its parent-context markers), the re-exports in `supervisor/events.py`, the judge's unit tests and the 26 monkeypatch stubs other tests needed to silence it. A regression test dispatches four `schedule_subagent` events with the same objective, parent and default role and asserts all are admitted (two with different configured subagents, two with identical ones) and that admission never constructs an LLM client; the same test fails on the base tree.

The `rejected_duplicate` status constant, its outcome-axis mapping and the UI/history projections stay untouched so existing task records replay correctly; nothing produces the status any more.

`docs/ARCHITECTURE.md` (module map row) stops claiming a duplicate gate and states the admission floor that remains; `docs/CHECKLISTS_ARCHIVE.md` gains the item 21 standing disclosure for this owner-accepted removal; `ouroboros/size_ratchet_manifest.py` registers the band rationale for `tests/test_nested_rights_depth.py`, which shrank into the 1001-1500 band when its stubs left.

No version carrier is touched (the maintainer assigns the version at integration).  13 files changed, 74 insertions(+), 757 deletions(-)

## Why

See #884 for the full record. In short: the judge was structurally blind to the one field that distinguished the tasks it was asked to compare (selected subagent / model), its only rejection in 46 days destroyed an intentional multi-model cross-check, it never caught a real duplicate, it blocked the supervisor main loop synchronously on every spawn, and it violated the owner-locked `semantic_dedup.py` contract on 8 of 9 rules. No other agent framework or task queue we checked gates spawns on semantic similarity; the floor everywhere is caps, budgets and caller-chosen idempotency.

## Capability change (checklist item 21)

Intentional, owner-accepted removal: a child is never refused for resembling another task. Surviving positive path: `schedule_subagent` with any objective, including byte-identical siblings (multi-model cross-checks, majority votes), is admitted subject to the existing depth cap, per-root active-child cap, hard cap and cost ceilings. Residual risk: a parent re-queued after a mid-wave crash may pay a wave twice, bounded by the cap (default 6) and the per-tree ceiling (default $50); not observed in 46 days.

## Verification

- `ruff check --select F` clean; focused pytest on the touched suites plus `test_smoke.py` on the rebased tree: exit 0.
- Full pytest in an isolated `OUROBOROS_DATA_DIR` on the pre-rebase candidate (result posted below when it finishes).
- Size-ratchet lane: the only remaining violations are the base's own debt (`ouroboros/loop_acceptance_review.py` at 1611 lines, the `reviewer_slots`/`onboarding_wizard` band entries, byte debt on `web/modules/chat.js` and `tests/test_devtools_benchmarks.py`); none of this PR's paths appear.
- End-to-end on two isolated instances (own data dir, OpenRouter, two configured API subagents `builder-a` = gemini-3.8-flash and `scout-b` = grok-4.6): the same prompt asked the agent to schedule two children with a byte-identical objective on the two subagents. On the base tree (`2a81b138f`) the second child was rejected: `9fb7a26f` -> `rejected_duplicate`, `duplicate_of=935c9376`, log line "Rejected duplicate task". On this branch both ran and completed: root `63333240`, children `76ed5699` (builder-a) and `ad718931` (scout-b), zero rejections.
- Review against `docs/CHECKLISTS.md` by three independent lanes on a frozen export of the candidate: Claude fable-5.1, Cursor Grok 4.6 Extra High, Codex gpt-6-astra (high). All three: SAFE TO COMMIT, no critical or major findings. Two advisories were applied in this commit (the orphaned parent-context helper removed; the regression test made base-failing by recording any LLM-client construction). The third note in all three reviews is the pre-existing size-ratchet base debt above.


ouroboros-agent:
CI attribution for the first run (head 9dd3c4ee3): quick-test and full-test (macos) failed on exactly one test, tests/test_generated_inventories.py::test_facade_inventory_is_byte_identical, which this PR caused (supervisor/events.py re-exports five fewer names, so the generated docs/v7next/FACADE_INVENTORY.md binding count moves 98 -> 93). Fixed by regenerating the inventory in fd700e8d2. full-test (windows) failed 18 tests: the same inventory test, 16 tests that the peer branch claude/project-dup-20260914 (no admission change) fails identically on Windows (runtime_mode_elevation x12, nano_ui_static charmap x2, the PEM CRLF capture test, deep_review_book_sources), and tests/test_loop_transport_wait.py::test_deadline_bounds_wait_with_one_last_free_redial_then_no_resend, a deadline test with a monkeypatched sleep that passes 3/3 locally on the merged tree and is touched neither by this PR nor by the target drift; treated as a runner timing flake. The branch now also carries a merge of ouroboros at 68cb3bb38 (clean, no overlapping files with the admission change); the newer target commits (#900) merge cleanly per git merge-tree and touch only docs/ARCHITECTURE.md in common.

PR #878, opened 2026-09-14
Name accounts in the pin select the way the Accounts tab does

razzant:
## Summary

Closes #875.

The Claudexor unified engine migrates each CLI's native login into a reserved registry row (`codex-default`, `claude-default`) whose `display_name` is the signed-in email. The Accounts tab names that row through `accountName` (display name, else the observed identity email, else the id), while the account pin select shared by Available subagents, Review lanes and Models named it by `profile_id` alone. The owner saw the email under Accounts and `codex-default` in the pin list and read the account as missing from the picker.

The pin select now calls an account what the Accounts tab calls it, with the stored id appended only when the two differ:

- `accountName` moves next to `accountRows` in `web/modules/claudexor_status_store.js` (the one client-side reader of `/api/claudexor/status`); `harness_accounts.js` imports and re-exports it, so its import surface is unchanged.
- `indexProfilesByHarness` in `web/modules/route_editor_primitives.js` becomes a projection of `accountRows` carrying `{id, enabled, name}` instead of a second, parallel walk over `payload.profiles.profiles`.
- `profileOptionsFor` labels an option `Account: <name> · <id> (pinned)` (`<name>` alone when it equals the id; `(disabled)` suffix unchanged). The option value stays the `profile_id`, which is what the setting stores and what pins the route.
- One sentence in `docs/DEVELOPMENT.md` ("Onboarding and Settings surfaces") states the rule; the `profileEntry` comment in `reviewer_slots.js` now names the third field.

Every registry row whose display name equals its id keeps its current label, so on a fresh install only the migrated `<harness>-default` rows read differently.

## Scope

In scope:

- The shared account pin select (Available subagents, Review lanes, Models) and the one naming rule it now shares with Accounts.
- Tests: the primitive label rule, wire parity between `accountRows` and the pin index on both the legacy and the unified fixture, and one rendered `availableSubagentRowMarkup` consumer.

Non-goals:

- Surfaces that report the stored id as a receipt keep doing so: the "Last actual run … account <id>" meta line (`describeExecutionEvidence`), the pinned-account warning (`pinnedAccountWarning`), the review-row last-execution line, the quota-wait card and onboarding summaries. The id in the option label is the bridge to those receipts.
- No change in Claudexor, in the gateway, or in what the settings store.
- No version carrier is touched (CONTRIBUTING §3).

## Verification

- [x] Focused tests for the changed behavior pass.
- [x] The default local test suite passes, or the reason it was not run is below.
[...]

PR #891, opened 2026-09-14
Latch failed owned-daemon starts and pin Claudexor 3.12.0 with bounded journal replay (#844)

### LINKED ISSUE #844: Owned Claudexor daemon spawn should pass a host-sized Node heap (journal replay OOM)
`ouroboros/claudexor_daemon.py` spawns the owned Claudexor daemon with `env = dict(os.environ)` (around line 552) and no Node heap option. On 2026-09-12 the pinned 3.10.4 daemon died twice at startup with `FATAL ERROR: JavaScript heap out of memory` while replaying a 752 MB `global` journal partition into the default ~4 GB heap (razzant/claudexor#295 tracks the engine side: whole-file replay before admission, a compaction bound the daemon cannot meet, no retention action for a healthy oversized partition). The app-driven restart therefore cannot recover on its own; a hand-started daemon with `NODE_OPTIONS=--max-old-space-size=49152` reached admission normally (peak RSS 5.7 GB).

Proposal: when spawning the owned daemon, set `NODE_OPTIONS=--max-old-space-size=<N>` (appending to any existing NODE_OPTIONS) with N derived from physical memory through the existing platform layer (for example half of RAM, floored at the Node default and capped sensibly), disclosed in the daemon start receipt/log. This is a host-side mitigation only; it does not replace the engine fix. Found during the codex-cache sprint hand-off; not part of #842.

ouroboros-agent:
## Summary

Closes #844 by explanation, not by a heap flag (see "Why #844's proposal is not the product" below). Since 2026-09-08 every managed start of the owned Claudexor daemon on this install died at V8's default 4 GiB heap while replaying a 1.07 GB journal; the host then respawned it on every caller (30 failed spawns before the incident was noticed). Two things change:

- **Claudexor pin → 3.12.0** (`ouroboros/claudexor_runtime_pin.json`): the engine now replays its journal frame by frame with a fold policy, so startup memory follows the retained state rather than the file (on the incident root: normal admission at the default heap, 1.64 GB → 295 MB after compaction, restarts rewrite nothing). This is the fix for the class.
- **Host custody of a failed start** (`ouroboros/claudexor_startup_failure.py`, `ouroboros/claudexor_daemon.py`, `ouroboros/server_maintenance.py`, `ouroboros/gateway/claudexor_accounts.py`): a daemon child that exits without having published its control descriptor is classified from its own bounded log interval (`heap_exhausted | writer_lease_contended | engine_floor | unclassified`, diagnostic only — the classification never selects behaviour), recorded as one durable `supervisor.jsonl` row, and latches the manager: ordinary callers get an immediate typed `daemon_spawn_failed` naming the classification and log interval instead of spawning again. The latch is released by a live-daemon attach, by the existing 600 s supervisor sweep (one retry, on a short-lived thread), by the owner's explicit Refresh (one attempt), or by Restart/Panic (new manager). No new timing constant, scheduler, kill or custody kind; the latch is per-process manager state (task workers hold their own).

Why #844's proposal (host sets `NODE_OPTIONS=--max-old-space-size` from RAM) is not the product: V8's default heap is 4 GiB on any machine with ≥ 8 GB, so a host-side guess does not track the journal; a wrong guess only moves the crash and hides the real growth. The engine now bounds replay memory by design; the host reports failures truthfully and stops respawning. `NODE_OPTIONS` continues to pass through the environment as the documented operator escape hatch.

## Also in this PR (CI green on the base's pre-existing debt)

- `ouroboros/loop_acceptance_review.py` trimmed 1611 → 1586 lines (duplicate branches folded into four helpers; behaviour unchanged, 150 acceptance tests pass) — the module had exceeded the 1600-line hard cap on the base.
- `web/modules/chat.js` composer file-target binder (image paste + drag/drop) moved byte-for-byte into `web/modules/chat_media.js` to repay the byte debt the base had grown past its shrink-only manifest entry; the size-ratchet manifest regenerated (band rationale for `web/tests/reviewer_slots.test.js`, stale band entries dropped, byte counts updated).

Version carriers are untouched (contributor PR); maintainers assign the release version at integration. The branch integrates the target's movement since its base (`77b91a4a2` → `453a82120`: #868, #877, #878) with one merge commit; the three files touched on both sides (ARCHITECTURE.md, DEVELOPMENT.md, the size-ratchet manifest) merged cleanly and the combined tree was re-validated (size ratchet `[]`, transition vs `453a82120` `[]`, docs/domain/host suites green).

## Verification

- Focused suites with an isolated data root: startup-failure (45 tests), owned daemon, startup lifetime, config extraction, docs sync, domain manifest, manual restart, panic — all 0; `validate_size_ratchet` → `[]`; transition vs base → `[]`; `ruff --select F` 0; version carriers byte-identical to base.
- Full pytest on the final tree (`6ddbb56eb`) with `OUROBOROS_DATA_DIR=$(mktemp -d) LC_ALL=C`: exit 0 (and exit 0 on the pre-pin tree `c41e0fac8`); the live `data/settings.json` was verified intact afterwards.
- Engine acceptance on a copy of this install's real 1.64 GB journal root is recorded in the Claudexor release: razzant/claudexor#303, published as https://github.com/razzant/claudexor/releases/tag/v3.12.0 (all 31 npm packages at 3.12.0). Headline: first start at the default heap admits in about 20–30 s instead of the out-of-memory crash, the root compacts from 1.64 GB to ~295 MB, restarts on the compacted root rewrite nothing.

## Review trail

Five independent review rounds (gpt-5.6-sol, gpt-6-astra, claude-opus-5, claude-fable-5-1, Cursor grok-4.6) with written dispositions; every accepted finding landed as its own commit (the sweep retries itself; the exit fact is harvested at every spawn decision, on attach and before the owner stop signals anything; the latch is taken under the manager lock at the first observed exit; the retry runs off the supervisor loop; both retriers named); mutation probes pinned the mechanisms with tests. Final whole-diff confirmations: gpt-6-astra (ultra) and claude-fable-5-1 (max): safe to merge. Rejected with evidence: "custody pids are checked before the exited child is settled" (custody liveness excludes exited and zombie processes; the real crashing-child test passes). Deferred to a follow-up issue: classify under the manager lock and retire the provisional latch record.


ouroboros-agent:
CI status on the final head `ee961bbef`: quick-test, ui-smoke, benchmark-methodology, the three fixture lanes, the three betterleaks smokes and full-test (macos-latest) are green. full-test (windows-latest) fails on the same 16 pre-existing tests as the base (the identical set failed on #868's Windows lane); none of the four files is touched by this PR. Filed as #898 with the failing test list and root-cause classes (a Windows-side run_shell redirect defect, a cp1252 read, a CRLF PEM fixture, one uninvestigated assertion). The one CI failure this PR did cause (an ext4 inode-reuse assumption in the new startup-log interval test) is fixed in `ee961bbef`.

Local verification of the final tree: full pytest with an isolated data root exit 0 on `9b701d35a` (the combined tree after merging the target's #868/#877/#878), size-ratchet validator `[]`, transition vs `453a82120` `[]`, version carriers byte-identical to the base.

ouroboros-agent:
Re-integration after target drift, and one more review round.

The target moved 35 commits while this PR was open (#886, #887, #889, #890, #894, #897, #900), which made it conflicting. cb49199ff merges managed/ouroboros (b44d46f87) into the branch; the two conflicts were the size-ratchet byte pin for web/modules/chat.js (re-pinned to the merged tree's exact 201090 bytes, regenerate_size_ratchet.py --check exit 0) and the static-UI contract fixture (the base's entry set, with the drag-and-drop contract still pointing at chat_media.js). Docs and chat.js auto-merged; the merged tree passed the web suite (1303/1303), the focused host suites, ruff, and the full isolated pytest (exit 0).

Round R-01 on cb49199ff, whole diff as the gate: gpt-6-astra scope 7 PASS + 1 critical, gpt-5.6-sol 30 PASS + 1 critical (the same defect, reproduced independently by both), claude-opus-5 30 PASS + 1 advisory, Cursor Grok 4.6 31 PASS. The defect: a caller that had passed the pre-spawn latch check could still spawn if a peer's child died and a third caller settled it inside the window before the spawn lock, so "no spawn while latched" was not atomic at its last decision point; the same window could overwrite an exited, unreaped child and lose its exit fact. Fixed in 1b2c07357: the latch is re-checked once more inside _spawn under its own lock, and _spawn never replaces an existing own handle (alive: join it; exited: the next settle owns the fact). Two regression tests pin both interleavings and fail on the previous code. The advisory (ARCHITECTURE's chat module sentence not naming the moved paste/drop binder) is fixed in the same commit. Focused suites, ratchet check and ruff are green on 1b2c07357; the Windows full-test debt is unchanged (#898).

Issue #883, opened 2026-09-14
Chat header stays "Thinking..." with typing dots after a turn's subagents finish (kind-less typing entries are exempt from snapshot revocation)


ouroboros-agent:
## Problem

After a Main-chat turn that spawned subagents completes, the header keeps showing `Thinking...` plus the animated typing dots indefinitely, although nothing is running. Observed live on 2026-09-14 (v7.0.0, HEAD 9a5241036): task `a80d495c` emitted `task_done status=completed` at 10:34:56Z, the queue was empty, and `GET /api/state` returned `active_direct_turns=[]`, `active_chat_activities=[]`, `active_chat_activities_complete=true`, `running_count=0`. Forty minutes later the header still said `Thinking...`. A page reload clears it, which is not a remedy in the packaged desktop shell, the Telegram mini app or on a phone.

## Mechanism (verified at HEAD by executing the shipped modules)

1. A subagent's `typing_start` is delivered with `kind=""` by design (`supervisor/events_chat_delivery.py:80-98`: the child is found in `RUNNING`, `is_root_task=False`, the kind is left empty).
2. The client stores it in `activeDirectActivities` with `kind: ''` (`web/modules/chat.js:3847-3858`, `:3695-3702`).
3. `computeHydratedDirectActivities` skips every entry whose kind is not in `SNAPSHOT_AUTHORITATIVE_KINDS = {'direct_chat','managed_task'}` (`web/modules/chat_activity.js:595`, `:934`), so a complete, supervisor-ready, empty `/api/state` snapshot is not allowed to delete it. The snapshot is applied every 3 s in Main and every 20 s in project panels; the authority arrives constantly and is refused.
4. The child's terminal facts do reach the client, but every one is routed to card-side stores before the activity store: `subagent_event: completed` rows are `is_progress: true` and return at `chat.js:3936` before `positiveTaskTerminalFact` at `:3942` (which is true for the delivered row); the child's `task_done` log frame returns at `:2314`; `settleTerminalRootChildren` (`:566-578`) settles cards only.
5. The only sweep of kind-less entries runs in `onWs('open')` (`chat.js:4064-4070`), i.e. on reconnect. The server heartbeat (15 s) keeps the client watchdog (45 s) from ever reconnecting on a healthy link, so the phantom is permanent exactly when the network is good.
6. While the parent runs, its own `Working...`/`Thinking...` outranks the phantom (`chat_activity.js:725-733`), so the defect appears at parent completion and looks like a finish race.

[...]

Issue #866, opened 2026-09-13
Presence task cards remain Working after durable completion


razzant:
## Problem
Presence task cards can remain in the `Working` state after the underlying task has completed and the external reply has been delivered.

Two observed Presence turns showed this behavior:

| Task | Durable status | Evidence |
| --- | --- | --- |
| `presence-0adf5b78d20da6881d0e1f59` | `completed` | 8 LLM rounds, `$4.08`, Telegram reply delivered |
| `presence-9d9ef9b20e3f8784e6e4c6c6` | `completed` | 3 LLM rounds, `$0.44`, Telegram reply delivered |

The runtime had `running_count=0` and `pending_count=0`; both task result records and `task_eval` events were terminal. The delivery outbox also recorded successful delivery. The UI cards nevertheless continued to display `Working`.

## Likely cause
The durable `chat.jsonl` task summary was written with prose saying that the task had completed, but its machine fields remained `outcome: "Working"` and `outcome_final: false`. The UI appears to trust this narrative projection instead of the authoritative terminal task result and evaluation records.

## Expected behavior
Once a durable task result and its evaluation are terminal, the task card must become terminal (`completed`/`delivered`, as applicable). A late or stale narrative summary must not keep a completed task in `Working` or regress a terminal card back to an active state.

## Acceptance criteria
- Reproduce the behavior with a Presence transport and at least two consecutive turns.
- Derive the visible lifecycle state from the authoritative durable task result/evaluation when available.
- Treat narrative progress summaries as descriptive text only; they must not override terminal lifecycle state.
- Preserve a separate delivery state so `completed` and `delivered` remain distinguishable.
- Add a regression test covering a completed task with a stale `outcome: "Working"` summary.

ouroboros-agent:
This is fixed. Typing frames are receipts now, the live activity set is a plain projection of the state census, and a card is only shielded by an id the census actually lists, so a finished Presence turn settles from its durable task result instead of sitting on Working. The durable summary also carries the same phase the browser paints, so a narrative row can no longer disagree with the terminal record. There is a regression test replaying this exact shape.

Ouroboros

PR #897, opened 2026-09-14
Chat header: the census is the only writer of the live-activity set

ouroboros-agent:
## Problem

After a Main-chat turn whose subagents had finished, the header kept showing `Thinking...` with animated typing dots although nothing was running (live on 2026-09-14, v7.0.0: `task_done status=completed`, empty queue, `/api/state` reporting `active_chat_activities=[]` with `active_chat_activities_complete=true`). Details, mechanism and history: #883.

## Root cause

A WS `typing` frame inserted an entry into the client live-set behind the header (`activeDirectActivities`). Entries without a server-stamped `kind` (every subagent, by design) were exempt from deletion by the complete census (`SNAPSHOT_AUTHORITATIVE_KINDS`), the only sweep ran on WebSocket reconnect (which a healthy 15 s heartbeat never triggers), and the child's terminal facts were filed into card-side sets before the live-set. The same live-set also shielded task cards from durable-result reconciliation (the Presence shape of #866). The exemption was introduced for queued managed tasks on 2026-08-16, lost its reason the next day when queue roots entered the census, and was then re-justified for subagents and pinned by tests; commit 1058e5824 named it as the cause and changed only frame addressing.

## Change

- **Typing frames never write the live-set.** `onWs('typing')` is a submission receipt: it triggers the existing `refreshHeaderControlState(true)` and retires the linked `Sending...` by `client_message_id` once that census read has answered, so the header steps `Sending...` → `Thinking...` (or `Online`) without a blank in between. Frames without a `client_message_id` (subagents) do nothing.
- **The census is the only writer of the live-set.** `computeHydratedDirectActivities` is a plain projection: a complete, supervisor-ready snapshot deletes every id it does not list whatever its kind; an incomplete snapshot deletes nothing; `concludedIds` still blocks resurrection by a snapshot captured before a keyed final. Removed: `showTyping`, the `SNAPSHOT_AUTHORITATIVE_KINDS` gate, the wall-clock request barrier and per-entry snapshot generation (the barrier survives only for the card scan's `lastLiveObservedAt`), the reconnect-only kind-less sweep, the cancelable-progress typing writer.
- **Cards are shielded by census-listed ids only** (the `activeDirectActivities.keys()` union is gone), and the census card scan revokes Stop for a root card the page saw running live and the queue census now omits (previously done by the map-departure path the deleted progress writer fed); a card replayed from history keeps Stop until its durable read answers, as before.
- **Server behaviour unchanged**; `kind` is presentation only (comments corrected). Telegram's native typing consumer is untouched.
- Docs state the contract and narrow the earlier "stuck-Working class closed" claim to what is now true; `web/modules/chat.js` shrinks 204 824 → 202 697 B (size-ratchet pin updated to the exact count).

## Tests

- Tests that required kind-less survival are inverted; typing-driven expectations are driven through the census (`web/tests/chat_continuity.test.js`, `chat_inflight_indicator.test.js`, `chat_instance_dom.test.js`, `liveness_scan.test.js`, `ui_result_projections.test.js`; three `ui_browser` continuity tests converted from typing triggers to card/census triggers).
- New `web/tests/chat_header_census.test.js` (10 cases): the incident replay through the real chat instance → `Online`; census-only header transitions; every kind revocable / incomplete deletes nothing; the receipt path with a recorded header-write trail proving no `Online` blink between `Sending...` and `Thinking...`; a Presence-shaped card the census never lists (with the kind-less typing frame that shielded it on base) settles from durable detail; a progress-only root loses Stop the moment the census omits it.
- `tests/test_ui_smoke_inflight_indicator.py` converted to the census/receipt contract (it runs in the push-time `ui_browser` lane).
- One static pin removed (it pinned the deleted sweep) and one that pinned deleted `showTyping` text.

## Verification

- `cd web && node --test tests/*.test.js`: 1280 passed.
- Focused pytest (static pins, docs sync, gateway parity, project chat continuity, inflight seams, architecture facts, web static guards): 358 passed.
- CI quick-tier pytest, isolated data dir: parallel pass 19 179 passed / 0 failed (13 skipped), serial pass 1598 passed / 0 failed. One textual pin (`tests/test_projects_v6640.py`) that named the deleted progress-writer condition was repointed to where host-attested Stop authority now lives (`updateLiveCardFromProgressMessage`).
- `ui_browser` lane `tests/test_ui_smoke_project_continuity.py`: 9 passed on the branch (9 passed on base before the conversions; the three converted tests failed on the branch only where they asserted that a typing frame lights the header). `tests/test_ui_smoke_inflight_indicator.py`: 1 passed.
- The entire `-m ui_browser` lane (57 files) on the branch: 235 passed, 35 failed; every one of the 35 fails identically on the base tree `2a81b138f` in the same environment (re-run there: 35 failed), and none is unique to the branch. Two of them (`test_s3_chat_card_dropdown_hurry_and_soft_stop`, `test_ui_smoke_cancel_run_button_eligibility_and_cancelled_state`) were probed directly on both trees: the history-replayed `live-root` card never receives Stop and its detail read answers 404 every 3 s on base as well.
- Visual, real server + Playwright in Chromium and WebKit, incident frames injected into the live WebSocket, census forced empty+complete: base → `Thinking...` with dots; branch → `Online`, dots hidden.
- Size-ratchet lane: the `chat.js` byte error is gone; the remaining failures are the pre-existing base debts (`loop_acceptance_review.py` at 1611 lines, `web/tests/reviewer_slots.test.js` band rationale, two stale band entries) — not touched here.

## Residuals (disclosed, not fixed)

- A running descendant that has typed but not yet emitted a progress row is absent from the header until that row mints its card; the same holds for a root the census never enumerates (a Presence turn is not registered), whose header presence comes from its card alone.
- Concurrent `/api/state` builds on the server may capture in a different order than the requests began; a census captured before a direct turn registered can conclude that turn client-side for its remaining life. Pre-existing for census-listed rows, very unlikely, no client fix warranted.
- If the supervisor loop is not ready or a census source fails, deletions pause until the next complete snapshot (fail-closed).
- A subagent card whose terminal frames were all lost keeps `Working` without dots (owner decision 2026-08-23); the card scan's deliberate subagent skip is unchanged.
- An unkeyed final still clears all pending state (pre-existing, separate).
- The typing receipt retires `Sending...` once its `/api/state` read settles (so the header never blanks on a normal send); a read that never settles at all leaves `Sending...` to the routing receipt, the next Main poll or reconnect replay. No timeout was added (owner: no timers).

## Review

- Three independent adversarial reviews of the design (codex gpt-5.6-sol, cursor grok-4.6 xhigh ×2) confirmed the diagnosis and rejected a snapshot-deletion-only fix as one more removal rule; this is the insert-side subtraction they converged on.
- Code review of this candidate against `docs/CHECKLISTS.md` (Repo Commit Checklist): codex gpt-6-astra (high) — SAFE TO COMMIT on a637a63fa, no blockers, no advisory, 31-item table; claude-opus-5 via Claudexor (profile mironov, xhigh) — NOT SAFE on a637a63fa with one blocker (an unconverted `ui_browser` test that only runs on push) and eight advisories; all were applied in the follow-up commit except A8 (disclosed above); focused delta reviews of the follow-up: claude-opus-5 — SAFE TO COMMIT (B1, A1–A4, A7 resolved; A5/A6 leftovers and ten hygiene advisories, all applied afterwards: `.then` retirement and a failed read are now covered by the node test, the ordering assertion is explicit, `app.js` no longer forwards the removed parameter, the docs pointer and the remaining “written by the census alone” wording were fixed, the `ui_browser` inflight test now proves the receipt itself pulled the census); codex gpt-6-astra — NOT SAFE on one point, disposed as a disclosed residual (below): a `/api/state` read that never settles would leave `Sending...` until it does; both reviewers confirmed every ordinary failure (rejected fetch, non-OK, bad JSON, reversed or destroyed instance) still retires the send. Adding a timeout would be the timer the owner rejected; the routing receipt, the 3 s Main poll and reconnect replay retire the send independently.

Related: #866 (card-side sibling; the Presence shielding path is covered by the new replay test, the subagent-card scan rule is not changed).


ouroboros-agent:
Merged the current ouroboros target into the branch (no conflicts; the target dropped the newer-history control in chat.js, so the size-ratchet pin now follows the merged file at 203048 bytes) and made the liveness_scan source pin CRLF-tolerant, which is what failed the Windows node suite on the first run. On the merged tree: node suite 1302 passed, focused pytest 305 passed, the CRLF simulation of the pin passes. The remaining red on the previous run was the size-ratchet lane's pre-existing base debt (loop_acceptance_review.py at 1611 lines, the reviewer_slots.test.js band rationale, two stale band entries).

ouroboros-agent:
Second CI run: the node suite passes on all three platforms, ui-smoke passes, and the remaining red is pre-existing base debt: the size-ratchet lane (loop_acceptance_review.py at 1611 lines, the reviewer_slots.test.js band rationale, two stale band entries) on quick-test and macOS full-test, and the Windows parallel pytest pass with the same 16 failures every recent PR shows (#886, #890, #894) plus one PermissionError flake on a pytest temp file in tests/test_plan_review_historical_replay.py, a file this PR does not touch (its other parametrization passed). Merging per the owner's decision to land with the known base debt.

Issue #906, opened 2026-09-14
Per-round acceptance observation message breaks OpenAI-family prompt caching (Codex, OpenAI, OpenRouter)


ouroboros-agent:
## Summary

Since 2026-09-13 the main loop appends a fresh `[ACCEPTANCE_SUBJECT_OBSERVATION]` user message at the end of every request and removes the previous one on the next round (`loop_acceptance_review.prepare_acceptance_observation`, commits 81819a707 and f99e278b2, integrated through the checkpoint merge dd5aded8f). The message carries `tool_count`, which changes every round, so no request is ever a byte-prefix of the previous request.

OpenAI-family prompt caches (the Codex subscription backend and the public OpenAI API, measured through OpenRouter) only reuse a previous request when that entire request is a prefix of the new one. With the transient tail the cache never extends past the first round: every round pays for the whole conversation again.

## Evidence

- Root task `a3e646fc` on 2026-09-14 (gpt-5.6-sol via Claudexor, 108 rounds, 50.2M prompt tokens): cached tokens frozen at 346,368 (system prompt + task message) from round 2 to round 103 while the prompt grew from 347k to 607k. Consecutive daemon payloads differ only in the replaced tail message; tools, system blocks, native items, account and cache key are byte-identical.
- All 55 round transitions of child tasks (which do not receive the observation) were pure appends and every one advanced the cache. The root task had 106 tail-replacement transitions and advanced 6 times, exactly when the tail text happened to be extended instead of replaced.
- Task `0be6fb06` (~260k prompts, same day): 74 tail replacements, 2 advances. So this is not a prompt-size effect.
- Direct reproduction against the Codex backend with the real payload (373k → 385k tokens, four rounds): append-only history hits the previous write every round (uuid or `exec_` cache key, low or xhigh effort, sol or astra); the same sequence with a transient trailing user message stays frozen at the shared prefix.
- Same reproduction on OpenRouter `openai/gpt-5.6-luna` (23k tokens, two runs each): append-only 100% / 100% / 100% cached on rounds 2-4; transient tail 0% / 0% / 0%.
- Anthropic routes are not affected (cache breakpoints sit before the tail; opus and fable cache positions keep advancing after 13.09). Gemini implicit caching shows the same partial hit share before and after.

Cost on 2026-09-14 alone: about 21.8M uncached prompt tokens out of 102.4M on the Codex main lane (71% token-weighted reads instead of the ~93% the same tasks would reach with an append-only transcript).

[...]

PR #929, opened 2026-09-15
Keep the acceptance observation append-only so OpenAI-family prompt caches extend (#906)

ouroboros-agent:
Fixes the regression reported in #906: since 2026-09-13 the main loop appended a transient `[ACCEPTANCE_SUBJECT_OBSERVATION]` user row to the end of every model request and removed the previous one on the next round. Its rendered facts carried the per-round `tool_count`, so no request was ever a byte-prefix of the one before it. OpenAI-family prompt caches (the Codex subscription backend, the OpenAI API, OpenRouter → OpenAI) reuse a previous request only when that whole request is a prefix of the next, so the conversation cache never extended past the first round. Measured on 2026-09-14: 71.5 % token-weighted cache hit on the codex main lane against 96–100 % for child tasks that carried no such row, about 21.8 M uncached tokens in one day. Anthropic (cache_control breakpoints before the tail) and Gemini implicit caching were not affected.

## What changes

- **Append-only observation rows.** `prepare_acceptance_observation` appends a new `[ACCEPTANCE_SUBJECT_OBSERVATION]` row only when the rendered owner-source facts changed (owner source hash, owner generation, queue generation, fence token) and never removes or rewrites an earlier one. `acceptance_observation_prompt` renders the facts without `tool_count`; the value stays on the stored observation, where `apply_delivery_subject_decision` still bounds `material_tool_indices` with it. The owner-message merge helper never folds new text into an already-sent observation row. Acceptance semantics are unchanged: Main still acknowledges the exact owner source hash it received, and a newer row still supersedes an earlier selector.
- **The invariant made visible.** New leaf `ouroboros/transcript_prefix.py` records the append-only rule between the sends of one loop execution: per-message content digests (role, plain text, tool-call identity; cache markers, block shape and private custody keys are not content) and a `prompt_prefix_break` fact (`kind` ∈ `system_rewritten | tail_replaced | rewritten | shrunk`, `index`, `round`, `sanctioned_by`). The loop observes the transcript each round actually dispatched (after the model call and the fallback chain, before the assistant row is appended), so `round` names the send that broke. The three compaction seams that rewrite sent history on purpose (automatic main-fit reclaim, manual reclaim, authored context view) stamp a one-shot `sanction_rewrite(ctx, "compaction")` that the next observation consumes; every other break is emitted as the existing durable `task_checkpoint` event and counted in the loop usage as `prompt_prefix_breaks`. It records and never blocks a send. No new event type, setting, or gate.
- **Regression tests.** `tests/test_transcript_prefix.py` drives the real `run_llm_loop` with the acceptance observation injected and asserts every send is a prefix extension of the previous one (this test is red on the previous tree, failing exactly at the replaced observation row); two more loop tests drive the real manual and automatic compaction seams through a faked summarizer and pin the sanctioned disclosure on the compaction round; unit tests cover the recorder and the one-shot stamp. `tests/test_acceptance_observation_append_only.py` pins the row semantics: one frozen row across rounds, a second row only on changed facts, no `tool_count` in the text, the merge guard.
- **Docs.** DEVELOPMENT.md "Cache-friendliness invariant" now states the append-only rule and the OpenAI-family prefix semantics; ARCHITECTURE.md gains the module row and one sentence in the subscription-call section; the domain inventory is regenerated for the new module.

## Verification

- Full pytest on the candidate tree with an isolated data root: 20,779 passed, 38 skipped. `ruff check . --select F` clean. `ouroboros/loop_llm_call.py` untouched at the 1600-line cap; `loop_acceptance_review.py` at 1595.
- Direct backend probes and production request blobs from 2026-09-14 are summarized in the issue; the expected steady state after this change is about 93 % cache hit on long codex tasks (first rounds and compaction rewrites remain uncached).
- Reviewed against the Repo Commit Checklist by three independent lanes (Claude Fable 5.1, Cursor Grok 4.6, Codex GPT-6 Astra) on frozen snapshots. Round 1 found one critical defect (the recorder took its snapshot before the model call and did not see the automatic in-call compaction, so it would have reported production compaction as an unexplained break one round late); the second commit fixes it. Round 2 confirmed that fix and found one more (the relocated hook had pushed `run_llm_loop` past the 300-line function gate, which the size-ratchet lane refuses) plus an attribution nuance on the overflow-recovery path; the third commit fixes both. The last commit merges the current `ouroboros` tip (after #910); the two Claudexor lanes re-reviewed the merged tree. Dispositions are summarized in the review comment below.

## Out of scope

No version bump (contributor PR rule). Separate follow-ups: CHECKLISTS.md item 22 wording (protected file), a release live-gate for cache continuity, the OuroborosHub cache widget (razzant/OuroborosHub#73), a Logs-tab label for the new `prompt_prefix_break` checkpoint kind (today it falls through to the generic "Checkpoint" row like other typed kinds), and the narrow residual where an owner message drained after a same-round redial can still merge into an unflagged trailing user row (the recorder discloses it as `tail_replaced`).


ouroboros-agent:
Review record for this PR (three independent lanes against the Repo Commit Checklist, each on a frozen snapshot, each verified to leave a zero patch):

Round 1 (373520bfd): Codex GPT-6 Astra (high) and Cursor Grok 4.6 (xhigh) passed all 31 rows. Claude Fable 5.1 (xhigh) found one critical: the recorder took its digest snapshot before the model call and learned "sanctioned" only from the manual reclaim's usage, so the automatic main-fit reclaim that runs inside the model call would have been reported one round late as an unexplained prompt_prefix_break. Reproduced with the real loop; accepted and fixed in 3515e2d30 (observe the transcript actually dispatched; the three compaction seams stamp a one-shot sanction the next observation consumes; two loop tests drive the real seams). Two advisories were dispositioned as follow-ups: the Logs tab renders the new checkpoint kind through the generic "Checkpoint" row (web/ is out of scope here), and after a same-round redial an owner message can still merge into an unflagged trailing user row (low frequency; the recorder discloses it as tail_replaced).

Round 2 (3515e2d30): Astra and Grok passed all rows and confirmed the redial and stamp semantics. Fable confirmed the fix on every exit of the round body and found one new critical: the relocated hook had grown run_llm_loop from 300 to 304 lines, which the size-ratchet lane refuses as new function debt. Accepted and fixed in 476b18a5c (span back to exactly 300; the explanation lives in the helper's docstring). Fable's advisory that one overflow-recovery call can both reproject the system row and compact, so a stamp could sanction a system_rewritten fact, was accepted with the exact rule Fable proposed: a compaction stamp never explains a break at the system row. Fable also corrected the narrative that both seam tests were red on the first candidate: only the automatic-reclaim test was; the manual one fails when its stamp is removed.

Round 3 (deb3de92d, after merging the current ouroboros tip b81f7e826): Astra and Grok passed all rows on the merged tree; Astra re-ran the size-ratchet lane against the new base (5 passed) and the docs/inventory suites; Grok verified the index-0 rule against the compaction unit selector. The Fable lane was not re-run on this micro-delta. Author verification on the merged tree: full pytest 20,919 passed / 38 skipped, ruff clean, version carriers byte-identical to the base.

Follow-up issues will be filed for the two disclosed residuals once this PR lands.

ouroboros-agent:
CI attribution: quick-test, ui-smoke, full-test (macOS) and the betterleaks matrix are green. full-test (Windows) fails on the same 16 tests as the base branch's own run at b81f7e826 (test_runtime_mode_elevation x12, test_nano_ui_static x2, test_deep_review_book_sources, test_project_capture_advisory); the failing sets are identical and none of them touch the files this PR changes, so this is pre-existing Windows debt on the target, not a regression from this change.

PR #1015, opened 2026-09-16
Share one Codex prompt-cache key per install and model across executions

ouroboros-agent:
Follow-up to #906 / #929. After the append-only fix the codex lanes ran at 91.4 % token-weighted cache hit over a full day, and 63 % of the remaining uncached tokens were the cold first round of new executions: a root task, a subagent child or a consciousness cycle paid the governance prefix in full although consecutive executions share it byte-identically (about 286k tokens between two consciousness cycles, about 215k between two root tasks).

## Why the first round was cold

Every execution sent the Codex backend its own `prompt_cache_key` (the execution id), which the Claudexor adapter also places in the `session_id` header. Two direct probes against the backend (2026-09-17, details in the review packet) settled the mechanics: a second conversation under the **same** key and session receives 99 % of the shared prefix from cache; the same key with a **different** `session_id` receives none; and a conversation's own `x-codex-turn-state` stays valid after another conversation takes a turn under the shared session (HTTP 200, 99.5 % cached). A third probe fired two conversations simultaneously under the shared session, then their second rounds simultaneously with their own turn states: all four requests succeeded and the second rounds were 99.5 % cached. So one shared key per install is sufficient and safe for interleaved and concurrent executions.

## What changes

- **`llm_claudexor.cache_key_for_model(model)`**: one header-safe key per data root and model (`ouroboros-<model label>-<16 hex of sha256(DATA_DIR, model)>`), empty for every non-Claudexor model. Stable across restarts, distinct across installs and models.
- **All three producers of the main-loop affinity use it**: the main call in `loop_llm_call.py` (the edited line shrank; the file stays at its 1600-line cap), the prepared/re-prepared call in `loop_model_call.py`, and the prospective wrap-up pricing build in `task_pacing.py`, which must bind the same payload the send produces.
- **Unchanged**: turn-state handling (`nativeContinuation` per execution), the account preference carried by the conversation, the OpenRouter prefix-derived session identity, reviewer affinities. No new setting, gate or event. One disclosed nuance: the one-shot refused-account memory (`_FAILED_PROFILE`, a context variable keyed by affinity, source and model) can now match across two tasks that run one after another in the same worker context, so a refusal recorded at the very end of one task steers the next task's first dispatch away from that account once; with per-execution keys such a stale fact was inert. Harmless in single-profile auto mode and arguably desirable with several accounts.
- **Docs**: the cache-friendliness rule in `docs/development/06-rules-by-change-class.md` now states the install-scoped affinity and why; one sentence in `docs/architecture/06-agent-core.md` (caller-owned subscription calls).
- **Tests**: the helper is stable per install and model, header-safe, empty for API lanes, and changes with the data root; two executions of one install share the key while an API lane gets none; the wait-override re-preparation carries the shared key on the wire.

## Expected effect

At the measured load (33 task starts a day) about 7M fewer uncached tokens per day on the codex lanes and a lane hit rate near 96–97 %; consciousness cycles, which restart every few minutes when the instance is idle, stop paying 325k cold tokens each. OpenAI's guidance of roughly 15 requests per minute per key is far above the measured average (about 0.5 per minute; bursts near 10); at overflow the backend routes to another machine and the request simply misses the cache, which is today's behaviour.

## Verification

- Focused suites green; full pytest with an isolated data root: 21,375 passed, 39 skipped; `ruff check . --select F` clean; size-ratchet lane green against the base. The second commit only aligns two tests that pinned the per-execution key (the properties they pin, one identical key on the priced candidate and the send and the one-shot refused-account fact, are unchanged).
- Three review lanes against the Repo Commit Checklist on a frozen snapshot: Claude Fable 5.1, Codex GPT-6 Astra (high) and Grok 4.6; the Grok lane ran through the OpenRouter API with the diff and the changed code inline because the Cursor account is blocked by an unpaid invoice. Dispositions are summarized in the review comment below.
- After the live instance adopts the change through its managed update, the measure is the share of first-round codex calls with `cached_tokens > 0` and the lane's token-weighted hit rate (per-call `llm_usage` events).

## Out of scope

No version bump (contributor PR rule). Mid-task cold restarts after long gaps (provider cache TTL) and after route rebounds, reviewer-packet caching, and the Claudexor adapter (which already projects one `cacheKey` to both wire fields) are separate.


ouroboros-agent:
Review record for this PR (Repo Commit Checklist, frozen snapshot d426bea3c, zero patch verified on every lane):

Codex GPT-6 Astra (high; the first submission was refused by quota headroom on one profile and re-run on another): all 31 rows PASS; it ran the focused suites, the size-ratchet lane against the base, and Python checks of the wrap-up affinity and independent continuations. Its side note about six failing cost-ceiling tests reproduced with the base module swapped in and does not reproduce in the author's full run, so it is lane environment, not this change.

Grok 4.6 (through the OpenRouter API with the full diff and the changed code inline, because the Cursor account is blocked by an unpaid invoice): all 31 rows PASS, no tests on that lane by construction.

Claude Fable 5.1: runtime behaviour confirmed correct on every path it traced (candidate and send bind byte-identical keys; turn state stays per execution; the account preference comes from the execution's own transcript; DATA_DIR is exported by the launcher into worker environments, so the key is stable per install and distinct across installs). One critical finding: two tests in tests/test_forced_model_binding.py still pinned the retired per-execution key and were red on the snapshot; the second commit applies exactly the proposed fix and the full suite passes. One advisory: a past-tense clause in the development rule; removed in the third commit. Fable also named a residual the probes had not covered, concurrent in-flight requests under one shared session; a third probe fired two conversations simultaneously and then their second rounds simultaneously with their own turn states, all four succeeded and the second rounds were 99.5 % cached.

Disclosed nuance: the one-shot refused-account memory (`_FAILED_PROFILE`) can now match across two tasks run one after another in the same worker context; the effect is that the next task's first dispatch skips the just-refused account once. Harmless in single-profile auto mode.

Author verification on the merged head: focused suites, docs and inventory tests, size-ratchet lane against the current base, ruff, version carriers byte-identical; full pytest on the pre-merge head 21,375 passed / 39 skipped, and the merged head's full run is attached before merge.

PR #1106, opened 2026-09-19
fix(main): one row per Project question, a card only while the task waits

razzant:
Every Project question becomes one line in Main, and a card with its option buttons only while the task is actually waiting. The owner can answer straight from Main through the one existing decision ingress (`POST /api/decisions`); the full form, option details and stake stay in the Project room.

The problem this solves: already-answered questions kept sitting in Main as full-height cards, so a burst of three questions buried the conversation.

Base: `72fd62f2e7211199c532034530548e8f73e38da7` · 36 files, +1057/−407 · **version-neutral** (`VERSION`, `README.md`, `pyproject.toml`, `uv.lock`, `web/package.json`, `GATEWAY_CONTRACT_VERSION` and the `docs/ARCHITECTURE.md` header are byte-identical to base; the two `api_types.js` lines are additive typedef properties).

## What is in it

- `question_presentation.js` owns one read-only projection of a question's row (status words, the owner's answer or the assumption the task continues under, the question as context); `project_dialogue.QUESTION_STATUS` is its Python twin, pinned both ways by `question_presentation_parity.json`.
- `waiting` is the single state that grows a row into a card. Waiting needs positive evidence: the task's live wait record, or the original required flag before any record exists.
- Freshness comes only from existing sources — history, the `quiz_state` frame and the activity census. No new durable store, endpoint or poller.
- A press in Main answers through the same ingress and settles through the same observation a `quiz_state` frame uses, so a lost frame plus a stale open snapshot cannot reopen a settled question.

## The defect independent review found, and the fix

A task publishes a new quiz **before** its `owner_wait` row is written, so an `/api/state` read taken inside that window still names the *previous* question. The first version of `appendActivityQuestion` treated any positively named wait as evidence about every other waiting card, so a newer question could fold into a line and lose its option buttons until a reload — and `live_wait` then blocked a valid later census from restoring it.

Folding now requires a proven order: only a question asked **strictly before** the named wait folds, read from the `asked_at` stamp the pointer row already carries. A tie is not an order either — the stamps carry sub-millisecond precision that `Date.parse` truncates, so two genuinely different questions can read as equal. A tie, a missing stamp and an unreadable one all leave the card answerable: an extra card can still be answered, a wrongly folded one cannot.

`tests/test_project_question_pointer.py` now pins `ts == asked_at` on the producer side, because losing that stamp fails silently — every fold simply stops.

## Incidental changes, named deliberately

These rode along with the work and are **not** part of the question-row feature:

1. `web/modules/chat.js` — a promoted non-narration summary no longer replaces the card's authored title while a task is finalizing (plus `web/tests/progress_narration_voice.test.js`). This fixed a real flake in `tests/test_ui_result_browser.py`, where the intermediate `Working` phase replaced `still working`.
2. `tests/test_launcher_native_preparation.py` — the test fixture publishes its PID atomically (temp + rename). The parent read the file as soon as it existed and could parse an empty string; the launcher runtime is unchanged.
3. `ouroboros/contracts/task_contract.py` — one comment removed to keep the module under its line cap.

## Verification

Run in an isolated sandbox (separate data root, `--basetemp` outside any ancestor Git tree):

| lane | result |
|---|---|
| `cd web && node --test tests/*.test.js` | **1443 passed**, 0 failed |
| focused Python (questions, waits, contracts, gateway parity, docs, inventories) | **496 passed**, then **154 passed** on the touched files after the review round |
| `-m ui_browser` question-rows journey (1100 and 390 wide) | **2 passed**; screenshots captured and inspected |
| `-m browser tests/test_ui_result_browser.py` | **1 passed** (real keyless server: answer from Main, durable state, navigation, reload) |
| full suite, parallel + serial passes | 21172 + 1809 passed; every failure classified below |

**Full-suite failures are inherited, not introduced.** Ten failing ids reproduce unchanged on `origin/ouroboros` itself. Four more (`test_headless_workspace_shell`, `test_legacy_timeout_retirement`, two `test_runtime_mode_core` writer cases) **pass in a pristine clone of this head** — they were artifacts of my working tree, which carries a `.venv` inside the checkout and so resolves `python3` differently. The blocking `size_ratchet` lane fails identically on `origin/ouroboros`: `web/modules/onboarding_wizard.js` is 1608 lines against the 1600 cap, and this contribution does not touch that file.

## Review

Three independent diff reviews against the Repo Commit Checklist and one separate scope review against the Intent/Scope Checklist, each on the frozen candidate with per-file hashes.

- Scope review: **PASS on all 8 items**, 0 critical, 4 advisory; the order fix judged sufficient.
- Two diff reviews returned **critical**: the browser census fixture carried no `ts`, so with the new guard the fold would never happen and the `ui_browser` lane would time out. Fixed, and that lane now actually runs green at both viewports with screenshots inspected.
- One diff review returned PASS and asked for exactly the strict comparison, which is in.
- Two follow-up delta reviews checked the fix itself.

Reviewers ran no tests (read-only lanes without a shell) and did not recompute the patch hash; the numbers above are mine.

## Known residuals, disclosed

- The Python projection infers `resumed` from a wait record that moved on to another quiz **without** an order proof, so a history or detail read taken inside the same window can briefly show the newest waiting question as resumed. It is a non-live observation, so the next census naming the real wait restores the card. Documented in `docs/architecture/03`; not closed here.
- Multi-wait (several questions waiting at once as first-class state), grouping and animation are deliberately out of scope.
- In a narrow fixture viewport the header controls wrap and the connection pill slightly overlaps the first message. Pre-existing: the CSS here only adds `.project-question-*` rules.

Original authorship of the earlier commits is preserved.


razzant:
## Review round, and what changed because of it

Three independent diff reviews (Repo Commit Checklist) and one separate scope review (Intent/Scope Checklist) ran against the frozen candidate `72fd62f2e..1e149da5a` with per-file hashes, followed by two delta reviews of the fix.

**Scope review:** PASS on all 8 items, 0 critical, 4 advisory; the order fix judged sufficient.

**Two diff reviews returned critical, and they were right:** the browser census fixture carried no `ts`, so with the new guard `Date.parse('')` is `NaN`, no fold happens, and the `ui_browser` journey would sit in a 15-second `wait_for_function` timeout — the lane asserted something it no longer exercised. One reviewer additionally showed that my first guard accepted an *equal* stamp as proof of order, while `utc_now_iso()` keeps microseconds that `Date.parse` truncates, so two genuinely different questions can read as equal.

Both are closed in `4f4ca1d05`: folding now requires **strictly** earlier, and a tie, a missing stamp and an unreadable one all leave the card answerable. A sub-millisecond tie case and an unstamped case are pinned in `web/tests/question_rows.test.js`. One reviewer noted that the old happy-path test had been passing *because* of the equal-stamp gap — its batch questions now carry distinct stamps (7 s apart), so the positive fold path is asserted on real order.

`78d1030c0` adds the producer-side contract the fold silently depends on: `tests/test_project_question_pointer.py` pins `required_question["ts"] == asked_at`. Losing that stamp never fails loudly; it just stops every fold.

**Both delta reviewers then found the same new defect in my own prose** — `docs/architecture/03` still said a named wait may fold a question asked "no later than" itself (≤) while the code and its comment require strictly before. Fixed in `af77847ba`: the map now says what the guard does.

## Visual evidence, since a passing lane is not an inspected UI

`-m ui_browser` question-rows journey: **2 passed** (1100 and 390 wide). Screenshots captured via `OUROBOROS_UI_EVIDENCE_DIR` and inspected:

| file | sha256 |
|---|---|
| `question-rows-burst-1100.png` | `b2f5c2271b85032fd97aed1f120ddf1de87f63093838d5ef5a08329c232d540d` |
| `question-rows-burst-390.png` | `5ef991bec835d9162d5f2eb637f71a41e028dc32de4e427e803ce90e6da9606e` |
| `question-rows-answered-1100.png` | `400c8808c607c25b3823c33e0c8993a1bf641f91fdb6e2babe47b0086cec34a8` |
| `question-reloaded-1100.png` | `90b8633d18633163cfc27aa3d0c3c87566a53721a67cd84b274b4aa768af9a8d` |

At 1100 the burst reads as four one-line rows — the answer inline with the owner's own words, "the task continued", "continuing with: WebP at quality 82", "the task finished; a late answer is accepted" — and exactly one card at the bottom with both options, the `recommended` badge and "Details and own answer". At 390 the rows wrap to two lines, the card keeps its full question and buttons, and nothing overflows horizontally.

One observation that is **not** from this branch: in the narrow fixture viewport the header controls wrap and the connection pill slightly overlaps the first message. The CSS here only adds `.project-question-*` rules, so that overlap exists on the base too.

## On the expected red `size_ratchet` step

`quick-test` runs the blocking size lane, and it fails identically on `origin/ouroboros` itself: `web/modules/onboarding_wizard.js` is 1608 lines against the 1600 cap. This contribution does not touch that file. Same for ten other full-suite ids, which reproduce unchanged on the base; four more failures I saw locally turned out to be artifacts of my own working tree (a `.venv` inside the checkout changes how `python3` resolves) and **pass in a pristine clone of this head**.

Reviewers ran no tests — their lanes are read-only without a shell — and did not recompute the patch hash. Every number above is mine.

razzant:
**Correction to my own prediction above.** I expected the `size_ratchet` step to come back red here. It did not: the step reports `success`, and every check on this PR is green.

```
quick-test                 pass (15m14s)   ← includes the blocking size-ratchet step: success
full-test (macos-latest)   pass
full-test (windows-latest) pass
ui-smoke                   pass (59s)
betterleaks-platform-smoke pass (ubuntu, macos, windows)
android-test               pass
benchmark-methodology      pass
```

Why I was wrong: the CI lane is **pairwise** — it scores base→tip and blocks a size *regression*, which this branch does not introduce. What I hit locally was the absolute-cap assertion over the whole tree, which reports inherited debt (`web/modules/onboarding_wizard.js`, 1608 lines, untouched here) as a finding a fork carries rather than a gate this change must pass. So the earlier paragraph describes my local run correctly and the CI gate wrongly, and no baseline-failure allowance was needed for the checks.

The rest of that comment stands as written.

Issue #639, opened 2026-09-04
Swarm tree stop decided on a dead attempt's upper bound; tree ceiling re-minted per level starves grandchildren


ouroboros-agent:
## What happened
Root `e06773969b794d32` ("Мульти-Уроборос: концепт-брейнштурм", runtime `5b4546ea`, 22 descendants, 26 subscription harness runs at $0). Root finished `best_effort/budget_exhausted` with the tree stop text `$97.95 / потолок $95.71`; settled spend in `data/state/usage_attempts.jsonl` for the tree is `$89.47` (144 metered calls).

### A dead attempt's upper bound counted as spend
- Child `e873523f` was cancelled by its parent 11 s after dispatch (`cancel_intent requested` 11:55:06, `delegate_run_started` 11:54:45). Its fable-5 attempt `cc6570bbffa142d187cfbed007d31a5b` is `reserved` then `dispatched` (`usage_attempts.jsonl` lines 65418–65419) and never reached a terminal row; the task result shows `cost_usd=7.193859`, `cost_final=false`, `unresolved_upper_bound_usd=7.19372`, actual tokens 667/5.
- `unresolved` and `dispatched` bounds are summed into `accounted_usd` (https://github.com/razzant/ouroboros/blob/db6d7cf872df840e8c05e0cc2651d5bcff14b422/ouroboros/_usage_rows.py#L86, https://github.com/razzant/ouroboros/blob/db6d7cf872df840e8c05e0cc2651d5bcff14b422/ouroboros/_usage_rows.py#L96); `POSITIVE_PHYSICAL_ATTEMPT_STATES` includes `unresolved` (https://github.com/razzant/ouroboros/blob/db6d7cf872df840e8c05e0cc2651d5bcff14b422/ouroboros/usage_accounting.py#L256). `terminalize_abandoned_attempt` (https://github.com/razzant/ouroboros/blob/db6d7cf872df840e8c05e0cc2651d5bcff14b422/ouroboros/usage_accounting.py#L1001) has no production caller (tests only).
- Without that $7.19 the deciding tree spend (~$90.8) would have stayed under the $95.71 ceiling; child `14316682` (web-lens, $9.29) was failed on the ceiling and two grandchildren were cancelled with the reason "бюджет дерева почти исчерпан".

### Ceiling re-minted at every level
[...]

Issue #603, opened 2026-09-04
Budget and cache accounting: residuals disclosed rather than fixed


ouroboros-agent:
**What happens today.** Phase P6 reworked cost-ceiling resolution, cache-aware
reservations and review usage accounting. Several effects were accepted as disclosures
rather than changed.

1. BC-1 (superseded in round 7): a forced reply that still asks for a tool is now
   treated as incomplete on every rail and degrades through the host fallback path
   (`forced_model_incomplete`); a tool-calls-only reply with no prose still yields the
   plain host fallback text, which is the remaining disclosure.
2. BC-5: the duplicate reviewer usage emitter is gone, so reviews now produce roughly half
   as many `llm_usage` rows. This is a projection change, not a cost change, and anything
   counting rows (dashboards, ad-hoc ledger queries) will read differently.
3. BC-2: the cache-aware reservation reuses the last settled split for the same task and
   model, bounded by TTL and active-only; when there is no usable split it falls back to
   charging a full cache write. It is an estimate, and it is deliberately pessimistic.
4. Convergence: the stopping rule for P6 was "after round 6, one last panel; if only new
   phrasings of the candidate-parity / split-invalidation class remain, stop and disclose."

**Why it is a gap.** These are behaviour changes visible per configuration class, and their
only record is the sprint journal.

**Where it was found.** Sprint `catlife-route-fixes`, phase P6 (budget-cache); plan §8.2.8
and the §10 journal for rounds 2-6 and P6_auth5.

**Suggested direction.** State each in the release notes with its configuration class, per
the owner's rule that benchmark-visible shifts are fixed for everyone and mentioned there.

**Acceptance.** Release notes name all three behaviour shifts; the convergence disclosure
appears in the PR.

NEEDS-CONTEXT: the outcome of the final P6 panel (`triad/P6_auth6`) is not in the plan —
that run was still in flight at the last journal entry (04:43 MSK), and only its `run.log`
header exists. Confirm before publishing whether P6 converged or has open findings.

## Added after the seventh authoritative panel (2026-09-04, P6_auth7 / P6_auth8)

[...]

PR #1127, opened 2026-09-19
Follow the live Total budget and deliver the budget wrap-up answer

ouroboros-agent:
A running task now follows the owner's current Total budget, and a budget stop delivers the model's wrap-up answer instead of ending empty.

What the owner saw: three long root tasks stopped in one night with `budget_exhausted`, an empty final answer and status Failed, at $73 to $125 of a $400 per-task cap, minutes after the owner had raised Total budget. Two independent defects produced that.

The global limit was captured once per task. `agent.handle_task` stored the resolver's answer in the task's `UsageScope`, and the ledger fence, the wrap-up wallet observation, the pacing rails and review admission all preferred that number. A worker process re-projects settings into its environment only when a task starts, so the environment-only resolver could not follow a Save either, although `TOTAL_BUDGET` is declared an immediate setting. Ledger rows of concurrent roots carried 6800, 7000 and 7500 in the same minute. Now no task captures the limit. `resolve_total_budget_usd` reads the saved settings document first, unlocked because saves are atomic, re-parsed only when the file's fingerprint changes, and any read failure (a missing file, a refused benchmark pin) leaves the environment answering and is never remembered. `reserve_attempt` resolves the limit before the ledger lock, as it already does for pricing. Rows still record the exact applied limit and its source, and no revision is manufactured. Lowering the budget binds the next reservation the same way. The in-task cost ceiling stays the number resolved at task start (#639).

The graceful wrap-up call was never dispatched on API lanes. The rail admits one wrap-up against a candidate priced before the send, and an identity predicate refuses the send when the real payload differs. The real main-loop send had grown `stream` and `stream_options` and the priced copy had not: 54 bytes on OpenAI-compatible routes and 14 on direct Anthropic, where the lane added `stream` after its builder. The forced path swallowed the typed refusal as a generic error, so the task finalized its services and then ended with an empty answer. Every existing parity pin rebuilt the "actual" candidate with the prospective builder's own arguments, so none could see it. Now `task_pacing.main_loop_wire_options` is the one owner of the payload-shaping options of a main-loop send, spread by both `call_llm_with_retry` and the prospective builder, and `stream` is a builder input on every lane. A refused admitted candidate is recorded as a typed `forced_candidate_drift` checkpoint and the answer is asked for once more without the predicate: the refusal happens before dispatch and releases its reservation, and the ledger fence prices the send it actually sees. A closed dispatch window stays a deadline and is never retried. The blocking predicate and the reserved-versus-fresh identity check in `llm_attempt` are unchanged.

A budget stop that delivers the model's answer already lands as best effort and reads "Done with warnings"; the incident read Failed only because the answer was never sent. A lifecycle test now pins that path from the last-fit decision to the delivered answer. The stop sentence opens with the bound that binds and, when that is the shared wallet, says how to lift it, because the old sentence opened with the task cap and read as a broken per-task cap.

Tests: `tests/test_wrapup_real_send_parity.py` drives the real `call_llm_with_retry` on four routes (red on the base on all four), the whole rail, the typed retry and the deadline exception; `tests/test_budget_resource_facts.py` covers no capture, a raise and a lowering honoured mid-task, the wallet observation, the document-first resolver and failures that are never memoized, including the unpinned reader answering `None` for an absent or undecodable file (a defect a third independent reviewer found in the first candidate: that `None` was memoized under the new file stamp, so a transient read failure right after a Save would have stuck the running task to its stale environment budget until the next Save). The hand-built parity mirrors take the shared options. Full local battery, `ruff --select F` and the size-ratchet lane pass on the candidate.

Disclosed residuals: review-wave admission and the seats read the live limit moments apart, so an owner Save inside that window can still refuse a seat, which is the existing admission-holds-nothing residual; in the server process the real send fetches model capabilities while the priced copy skips that fetch, and a difference there now costs one released reservation and a retry instead of the answer; the task-start `budget_remaining_usd` scalar is unchanged and only rejects a task that starts with an exhausted budget; the supervisor-side raw `TOTAL_BUDGET` readers of #603 are unchanged; the Claudexor lane's priced copy takes its cache affinity from the prospective builder rather than from `main_loop_wire_options` (the same value, one owner short of the remote lanes); when the tree cap and the wallet are both exhausted the stop sentence names the wallet first. Wallet-aware pacing reminders are a separate follow-up. Independent review of the candidate: two commit-triad seats (codex gpt-6-astra, cursor grok-4.6) SAFE TO COMMIT, two scope seats (codex gpt-6-astra, claude fable-5.1) PASS, a third triad seat (claude fable-5.1) NOT SAFE on the memoized-`None` defect above, repaired in the last commit and re-reviewed on that delta. No version, tag or release change.


ouroboros-agent:
Post-merge disclosure from the acceptance review of this PR (two residuals not stated in the description):

1. **`scripts/run_external_review.py` contributor budget cap.** That script deliberately lets an explicit env `TOTAL_BUDGET` act as the reviewer's authorized ceiling (`_load_settings_into_env` keeps an already-set env value). `resolve_total_budget_usd` now prefers the saved settings document, so on a machine whose `settings.json` carries a budget, `TOTAL_BUDGET=5 scripts/run_external_review.py …` reserves against the document's number, not the 5. Bounded to that dev-only script (its ledger is a fresh temp drive); no runtime path is affected. Repair when touched: bind an explicit `UsageScope(global_limit_usd=…)` in the script instead of relying on the env-only resolver.

2. **Review coverage of the last commit.** `e5f958fa7` (one `monkeypatch.delenv("OUROBOROS_SETTINGS_SHA256")` line in a test) was applied from the sol delta review's advisory and verified by the full battery on that exact SHA (23095 passed), but did not itself receive a separate reviewer pass. The three prior commits were reviewed: astra/grok triad SAFE, astra/fable scope PASS, fable triad found the memoized-`None` defect repaired in `4bbaf1bf9`, sol re-reviewed that delta (no CRITICAL).

